PR #32, opened 2016-06-17
light edits and some re-ordering of text and examples

vhorne:


JMesser81:
looks good to me. Would be good if @grcusanz  would confirm and merge

PR #35, opened 2016-06-22
iDNS changes

Krash0x35:


SDNExpress/scripts/FabricConfig.example.psd1
```diff
@@ -40,6 +40,19 @@
...
+            iDNSCredentialResourceId = "c6abefg6-44fb-45f6-89ec-5ebd890a144f"            
+            iDNSAddress= '10.60.34.9'                                      #Example: "10.0.0.7"
+            iDNSZoneName = 'contoso.local'                                    #Example: "contoso.local"
+            DnsPort=53                                                        #Standard DNS port. DO NOT CHANGE            
+            DnsIPAddress="169.254.169.254"                                    #Fixed IP address understood by DNS proxy. DO NOT CHANGE
+            DnsProxyServiceName="DnsProxy"                                    #Name of the service. DO NOT CHANGE
+            iDNSMacAddress = 'AA-BB-CC-AA-BB-CC'                                  #Example "AA-BB-CC-AA-BB-CC"
+
+            #iDNS Proxy Registry Hives
+            iDnsVfpPath = "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\NcHostAgent\Parameters\Plugins\Vnet\InfraServices\DnsProxyService"
```
nwoodmsft: I don't think we should be exposing a registry path in the configuration file. This doesn't seem like something which is actually configurable? If this is a fixed path and it is needed by SDNExpress.ps1 then it would be better to have it within SDNExpress.ps1 itself (looks like only one Configuration called "ConfigureIDnsProxy" needs this registry path).

SDNExpress/scripts/FabricConfig.psd1
```diff
@@ -40,6 +40,19 @@
...
+            iDNSCredentialResourceId = "c6abefg6-44fb-45f6-89ec-5ebd890a144f"            
+            iDNSAddress= '<< Replace >>'                                      #Example: "10.0.0.7"
+            iDNSZoneName = '<< Replace >>'                                    #Example: "contoso.local"
+            DnsPort=53                                                        #Standard DNS port. DO NOT CHANGE            
+            DnsIPAddress="169.254.169.254"                                    #Fixed IP address understood by DNS proxy. DO NOT CHANGE
+            DnsProxyServiceName="DnsProxy"                                    #Name of the service. DO NOT CHANGE
+            iDNSMacAddress = '<< Replace >>'                                  #Example "AA-BB-CC-AA-BB-CC"
+
+            #iDNS Proxy Registry Hives
+            iDnsVfpPath = "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\NcHostAgent\Parameters\Plugins\Vnet\InfraServices\DnsProxyService"
```
nwoodmsft: Same comment as FabricConfig.example.psd1 - lets move these fixed registry paths to the ps1 rather than exposing in the configuration file.

SDNExpress/scripts/SDNExpress.ps1
```diff
@@ -2401,6 +2401,215 @@ Configuration ConfigureHostAgent
     }
 }
 
+Configuration ConfigureIDns
+{
+    
+	Import-DscResource -ModuleName 'PSDesiredStateConfiguration'
+
+	Node $AllNodes.Where{$_.Role -eq "NetworkController"}.NodeName
+
```
nwoodmsft: There are a bunch of blank lines and alignment issues in this Configuration. Can you please remove the blank lines and ensure that you are using spaces (not tabs) for indentation. This is for consistency with the rest of the .ps1 file.

SDNExpress/scripts/SDNExpress.ps1
```diff
@@ -2401,6 +2401,215 @@ Configuration ConfigureHostAgent
...
+        
+		Script CreateiDnsCredentials
+
+		{
+       
+			SetScript = {
+
+	                . "$($using:node.InstallSrcDir)\Scripts\NetworkControllerRESTWrappers.ps1" -ComputerName $using:node.NetworkControllerRestName -UserName $using:node.NCClusterUserName -Password $using:node.NCClusterPassword
+
+                    	write-verbose "Adding iDNS server credentials. UserName=$($using:node.DomainJoinUsername) Password=$($using:node.DomainJoinPassword)"
```
nwoodmsft: I'd need to check with MASD, but I think there was an effort to eliminate passwords being output into the script logs. It's ok to log that we are adding iDNS server credentials, but lets leave the actual creds out of the logging.

SDNExpress/scripts/SDNExpress.ps1
```diff
@@ -2401,6 +2401,215 @@ Configuration ConfigureHostAgent
     }
 }
 
+Configuration ConfigureIDns
+{
+    
+	Import-DscResource -ModuleName 'PSDesiredStateConfiguration'
+
+	Node $AllNodes.Where{$_.Role -eq "NetworkController"}.NodeName
```
nwoodmsft: This doesn't seem right - this means that you will execute these REST PUTs against every controller VM. I assume that you just want to execute a REST PUT once against the controller? If so, you should use:

SDNExpress/scripts/SDNExpress.ps1
```diff
@@ -2401,6 +2401,215 @@ Configuration ConfigureHostAgent
...
+			TestScript = {
+
+			Write-verbose "Executing Test iDns Configuration  Script"
+
+
+			. "$($using:node.InstallSrcDir)\Scripts\NetworkControllerRESTWrappers.ps1" -ComputerName $using:node.NetworkControllerRestName -UserName $using:node.NCClusterUserName -Password $using:node.NCClusterPassword
+
+
+			$iDnsObj = (Get-iDnsConfiguration)
+			#$zone = $iDnsObj.properties.zone
```
nwoodmsft: Remove commented out code

SDNExpress/scripts/SDNExpress.ps1
```diff
@@ -2401,6 +2401,215 @@ Configuration ConfigureHostAgent
...
+
+                # Enable firewall rules for DNS proxy service
+                Write-verbose "Start DnsProxy service and make it automatic"
+                Enable-NetFirewallRule -DisplayGroup 'DNS Proxy Service'
+
+                # Start DnsProxy service and make it automatic
+                Write-verbose "Start $($using:node.DnsProxyServiceName) service and make it automatic"
+                $dnsProxyService = Get-Service -Name $using:node.DnsProxyServiceName 
+                Set-Service -Name $using:node.DnsProxyServiceName -StartupType Automatic
+                Restart-Service -Name $using:node.DnsProxyServiceName
```
nwoodmsft: Do you want a -force here?

SDNExpress/scripts/SDNExpressTenant.ps1
```diff
@@ -171,7 +171,7 @@ Configuration CreateTenantVMs
             <Interfaces>
```
nwoodmsft: Fix alignment - please use spaces for indentation, not tabs.

SDNExpress/scripts/SDNExpressTenant.ps1
```diff
@@ -246,8 +246,15 @@ Configuration CreateTenantVMs
                     if ($($Using:node.productkey) -ne "" ) {
                         $key = "<ProductKey>$($Using:node.productkey)</ProductKey>"
                     }
-
-                    $finalUnattend = ($templateunattend -f $($Using:vminfo.ipaddress), $($Network.subnets[$using:vminfo.subnet].mask), $($Network.subnets[$using:vminfo.subnet].gateway), $($Network.DNSServers[0]), $($Using:vminfo.vmname), $($Using:Node.VMLocalAdminPassword), $key)
+		     if ($Using:node.UseIDns -eq $true) {
+                     write-verbose "Using iDNS"
+                     $finalUnattend = ($templateunattend -f "", "", "", "", $($Using:vminfo.vmname), $($Using:Node.VMLocalAdminPassword), $key, "true")
```
nwoodmsft: Are you 100% that this works? I would have thought that by using "" instead of actual values as a replacement for some of the placeholder in the unattend, we would end up creating an invalid unattend file. For example, wouldn't this lead to an unattend file with the following (IP specified as "/" instead of a proper address + mask):

```
                    <UnicastIpAddresses>
                        <IpAddress wcm:action="add" wcm:keyValue="1">/</IpAddress>
                    </UnicastIpAddresses>
```

Unattend usually isn't very forgiving about invalid entries and it could lead to either the tenant VM not booting or the tenant VM showing the OOBE instead of booting straight to a finalized OS. Please confirm.

SDNExpress/scripts/TenantConfig.psd1
```diff
@@ -29,8 +29,9 @@
 
             #Password to assign to the local administrator of created VMs
             VMLocalAdminPassword = '<< Replace >>'                              #Example: "V3ryC0mplexP4ssword"
+
```
nwoodmsft: The comment should probably go above the UseIdns = $true line, as it's explaining what the feature is, not providing an example.

PR #48, opened 2016-10-10
new resources for VMM SDN 

manishmsft:


devalian:
i was having difficulty to deploy SLB using these templates. i hope the latest commit can make it happen. :)

PR #61, opened 2016-11-22
updating two NC scripts

manishmsft:
This is an update about SFPCopy that didn't go through in the last attempt.

### DIFF
--- a/VMM/Templates/NC/NCSetup.cr/PrepareNodeForNetworkController.ps1
+++ b/VMM/Templates/NC/NCSetup.cr/PrepareNodeForNetworkController.ps1
@@ -25,25 +25,6 @@ try
     #------------------------------------------    
     Log "Installing NetworkController Role.."
     Add-WindowsFeature -Name NetworkController -IncludeManagementTools
-
-
-
-
-    #------------------------------------------
-    # Disable Restart Manager in NC for Windows Server RTM
-    #------------------------------------------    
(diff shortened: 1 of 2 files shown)


VMM/Templates/NC/NCSetup.cr/PrepareNodeForNetworkController.ps1
```diff
@@ -25,25 +25,6 @@ try
...
-    #------------------------------------------
-    # Disable Restart Manager in NC for Windows Server RTM
-    #------------------------------------------    
-    $manifestPath = ([System.IO.Directory]::GetParent([System.Environment]::SystemDirectory)).ToString()
-    $manifestPath += "\NetworkController\TemplateClusterManifest.xml"
-
-    [xml] $ncClusterManifest = Get-Content $manifestPath
-    $ncClusterManifestVersion = $ncClusterManifest.ClusterManifest.Attributes["Version"].Value
-
-    if($ncClusterManifestVersion -eq "10.1.0.0")
```
JMesser81: Are you sure we should remove this check? No one should be on pre-9D patch but if they are, this check helps catch it.

manishmsft:
yes, we can go ahead with this change. Customers will anyway encounter the failure with pre-9D build. It's the error message that would differ. This change has no impact on customers using 9D+ build.

PR #77, opened 2017-01-23
Update install script to allow overwriting parameters.

gkudra-msft:
The latest version of the plugin introduces support for public cloud setups. This is the behavior by default, this change adds parameters to change the behavior (this needs to be used in the private cloud stack).


gkudra-msft:
Things got a little messed up with permissions, hence the accidental double commit.

JMesser81:
LGTM

PR #97, opened 2017-05-04
Added version and hash to output

tmolenh:
Will output a txt file with the defined $version number and the SHA256 hash of the script itself.

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -1,5 +1,5 @@
-﻿
 
+$version = "17.05.0"
 
 
 <# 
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -1618,6 +1634,7 @@ function Worker {
     EnvDestroy        -OutDir $workDir
     EnvCreate         -OutDir $workDir
 
+    Version           -OutDir $workDir
```
omarcardona: Version output here means, its not visible on screen as the data scrolls by.  You should take the contents of Version() and execute them from inside Completion().  Such that it displays to screen at end of execution and to the output file you created.

tmolenh: Done, here's an output sample:

```
Diagnostics Data:
-----------------
C:\Users\<username>\Downloads\Get-NetView.PS1
Version: 17.05.0
SHA256:  CD73FCBC960EDB33A2F7C36263B80F702A6AF02A785860D7D208B94FB25D03F4

C:\Users\<username>\Desktop\msdbg.<computername>-2017.05.09_06.02.52.zip
...
```

PR #101, opened 2017-05-17
Added ability for user to specify output directory

tmolenh:
User can use `-OutputDirectory` to change the output location. ~~If the passed value is null, not specified, or doesn't exist, then it will default to the Desktop (a warning will be displayed in the last case).~~

**Update**
Will now only default to the Desktop if `-OutputDirectory` is not specified (or empty). If the directory does not exist or the user doesn't have write permission, an Exception will be thrown. This is intended to be consistent other Cmdlet behavior.

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -1,5 +1,8 @@
+Param(
+    [parameter(Mandatory=$false,HelpMessage="Complete Path to the Output Directory")] [String] $OutputDirectory
+)
 
-$version = "17.05.0"
+$version = "17.05.1"
 
 
 <#
@@ -1571,7 +1574,6 @@ function EnvCreate {
     New-Item -ItemType directory -Path $OutDir | Out-Null
 } # EnvCreate()
 
-
 function CreateZip {
     [CmdletBinding()]
     Param(
@@ -1626,11 +1628,28 @@ function Completion {
 } # Completion
 
 function Worker {
+    [CmdletBinding()]
+    Param(
+        [parameter(Mandatory=$false)] [String] $OutputDirectory
+    )
+
     clear
     $columns = 4096
 
+    if ($OutputDirectory) {
+        if (Test-Path $OutputDirectory) {
+            $baseDir = (Resolve-Path $OutputDirectory).Path
+        } else {
+            Write-Warning "Output path '$OutputDirectory' not found, defaulting to Desktop."
+        }
+    }
+
+    # Default to desktop directory
+    if (-Not ($baseDir)) {
+        $baseDir = [Environment]::GetFolderPath("Desktop") + "\"
+    }
+
     $workDirPrefix = "msdbg." + $env:computername
-    $baseDir       = [Environment]::GetFolderPath("Desktop") + "\"
     $workDir       = Join-Path $baseDir $workDirPrefix
 
     EnvDestroy        -OutDir $workDir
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -1605,53 +1607,70 @@ function Completion {
...
 } # Completion
 
 function Worker {
+    [CmdletBinding()]
+    Param(
+        [parameter(Mandatory=$false)] [String] $OutputDirectory
+    )
+
     clear
     $columns = 4096
```
omarcardona: Move the -$version at the top of file to here.  Next to the locals.  No need to have it disjoint.

Diagnostics/Get-NetView.PS1
```diff
@@ -1605,53 +1607,70 @@ function Completion {
...
 
+    if ($OutputDirectory) {
```
omarcardona: All of this new code should be in a subfunction.  Try to keep the code cohesive, where a function encompasses one action/intent.  Plus it avoids visual complexity in the Worker().

Diagnostics/Get-NetView.PS1
```diff
@@ -1556,23 +1554,42 @@ function Environment {
...
+        } # else $OutputDirectory is undefined
+        return [Environment]::GetFolderPath("Desktop") + "\" 
+    } # NormalizeOutDir ()
+
+    # Calculate working directory
+    $workDir = Join-Path $(NormalizeOutDir) $WorkingDirectoryName
+
+    # Delete working directory (and all of its contents) if it already exists
+    If (Test-Path $workDir) {
+        Remove-Item $workDir -Recurse
```
tmolenh: I put everything here because I didn't like the idea of `Remove-Item $workDir -Recurse` being separated from the definition of `$workDir`, especially now that user input involved.

omarcardona: No, please keep create() separate from destroy(), merging the two introduces unnecessary maintenance complexities.  As an example, the previous implementation was self descriptive and no need to comment.  The current implementation is overloaded in that it does two things, both create and destroy which goes against cohesion principles, requires comments to understand, and implicitly does destroy/create encompassed in setup.  You've also introduced a nested function which is not ideal.  It's extremely important that this script remain simple and easy to maintain.  Lastly, I don't follow why this change is required due to $workDir?  It was an input parameter in the previous function, why is it a problem which warrants this refactoring?  It was already grouped with other script wide vars such as $columns and $version.

tmolenh: The problem is that `Remove-Item -Recurse` will indiscriminately remove a directory and all its contents. Not an issue before because the output dir was hardcoded to the Desktop. Imagine if something like `%USERPROFILE%\Documents` were passed to `EnvDestroy`? My logic is that putting input validation and execution in the same step reduces the likelihood of bugs leading to data loss.

I'm not convinced that that `EnvSetup` is against cohesion because `EnvDestroy` and `EnvCreate` were overly simple in function and heavily dependent on one-another. There's never a reason to call one "by itself", i.e. without calling the other before or afterhand.

Why is a nested function not ideal? (a lambda would be better here, however) In this case, I think it makes slightly easier to see what the input to `Join-Path` is going to be.

Also, why is documenting code with comments worse than with function declarations?

omarcardona: Comments aren't enforceable.  That's why the existing code biases towards simple self descriptive logic vs comments.  Ping me when you're around to sync up.

tmolenh: Pushed some changes (with --amend) that remove the nested function and clean up the params passed to Completion.

PR #104, opened 2017-06-07
Add admin requirement

tmolenh:
Adds a soft requirement for Admin privileges. If the script is run in a normal PowerShell window, an exception will be thrown. The requirement can be bypassed with the `-SkipAdminCheck` switch, in which case the script will run as before, with some commands failing if not in admin mode.

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -22,7 +22,7 @@
 .PARAMETER OutputDirectory
     Optional path to the directory where the output should be saved. Can be either a relative or an absolute path.
     If unspecified, the current user's Desktop will be used by default.
-.PARAMETER NoAdmin
+.PARAMETER SkipAdminCheck
     Not implemented
 .EXAMPLE
     .\Get-NetView.ps1 -OutputDirectory ".\"
@@ -57,8 +57,8 @@
     https://github.com/Microsoft/SDN
 #>
 Param(
-    [parameter(Mandatory=$false,HelpMessage="Path to the Output Directory")] [String] $OutputDirectory<#,
-    [parameter(Mandatory=$false,HelpMessage="Bypass Administrator requirements")] [Boolean] $NoAdmin#>
+    [parameter(Mandatory=$false,HelpMessage="Path to the output directory")] [String] $OutputDirectory,
+    [parameter(Mandatory=$false,HelpMessage="Skip check for Administrator privileges")] [Switch] $SkipAdminCheck
 )
 
 function ExecCommandText {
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -162,54 +162,54 @@ function ExecCommand {
...
-        $out  = (Join-Path -Path $dir -ChildPath $file)
-        [String []] $cmds = "Get-NetIPInterface -InterfaceAlias ""$name"" | Out-String -Width $columns",
-                            "Get-NetIPInterface -InterfaceAlias ""$name"" | Format-Table -AutoSize",
-                            "Get-NetIPInterface -InterfaceAlias ""$name"" | Format-Table -Property * -AutoSize | Out-String -Width $columns",
-                            "Get-NetIPInterface | Get-Member"
-        ForEach($cmd in $cmds) {
-            ExecCommand -Command ($cmd) -Output $out
-        }
+    $name = $NicName
+    $dir  = $OutDir
```
omarcardona: Looks like this was just formatting the indentation.  Please confirm.

Diagnostics/Get-NetView.PS1
```diff
@@ -227,138 +227,138 @@ function NetIpNic {
...
-            ExecCommand -Command ($cmd) -Output $out
-        }
+    $file = "Get-NetNeighbor.txt"
+    $out  = (Join-Path -Path $dir -ChildPath $file)
+    [String []] $cmds = "Get-NetNeighbor | Format-Table -AutoSize | Out-String -Width $columns",
+                        "Get-NetNeighbor | Format-Table -Property * -AutoSize | Out-String -Width $columns",
+                        "Get-NetNeighbor | Get-Member"
+    ForEach($cmd in $cmds) {
+        ExecCommand -Command ($cmd) -Output $out
+    }
```
omarcardona: This looks like a reordering of the commands.  What was the rationale for that?  Specifically, did you ensure the entire command block was moved or just the command name?  If the former, then there's a high likelihood that some of the commands are broken.  On PS the parameters per command are super picky.  In my original implementation every single command was inspected to ensure correct execution/output.

Diagnostics/Get-NetView.PS1
```diff
@@ -1668,23 +1661,27 @@ function Completion {
...
-    Write-Host "`n$Src"
-    Write-Host "Size:   $dirssize"
-    Write-Host "Dirs:  "(Get-ChildItem $Src -Directory -Recurse | Measure-Object | %{$_.Count})
-    Write-Host "Files: "(Get-ChildItem $Src -File -Recurse | Measure-Object | %{$_.Count})
+
+    $dirs = (Get-ChildItem $Src -Recurse | Measure-Object -Property length -Sum) # out folder size
+    $hash = (Get-FileHash -Path $MyInvocation.PSCommandPath -Algorithm "SHA256").Hash # script hash
+
+    # Display version and file save location
+    Write-Host "`n",
```
omarcardona: Please preserve the old format.  Where individual items are separate Write-Host commands.  The original implementation was done for editing flexibility rather than runtime flexibility.  For example, one can apply modifiers to individual lines, whereas the new implementation is all or none.

Diagnostics/Get-NetView.PS1
```diff
@@ -1,35 +1,65 @@
...
-        Open the base directory with Visual Studio Code and review via the Navigation Panel
+.PARAMETER OutputDirectory
+    Optional path to the directory where the output should be saved. Can be either a relative or an absolute path.
+    If unspecified, the current user's Desktop will be used by default.
+.PARAMETER NoAdmin
+    Not implemented
+.EXAMPLE
+    .\Get-NetView.ps1 -OutputDirectory ".\"
+    Runs Get-NetView and outputs to the current working directory.
+.NOTES
```
omarcardona: Very nice!  I like this.

Diagnostics/Get-NetView.PS1
```diff
@@ -1594,7 +1610,7 @@ function GetWorkDir {
                    if (Test-Path $OutputDirectory) {
                        (Resolve-Path $OutputDirectory).Path # full path
                    } else {
-                       Throw "Get-NetView: The directory '$OutputDirectory' does not exist."
+                       Throw "Get-NetView : The directory '$OutputDirectory' does not exist."
```
omarcardona: Why was a space added between the word and ":"?

Diagnostics/Get-NetView.PS1
tmolenh: It's just the GitHub diff view going insane. Beyond Compare 4 correctly shows this as only whitespace changes.

Diagnostics/Get-NetView.PS1
tmolenh: It is (see below)

Diagnostics/Get-NetView.PS1
```diff
@@ -1596,7 +1641,7 @@ function EnvCreate {
     Try {
         New-Item -ItemType directory -Path $OutDir -ErrorAction Stop | Out-Null
     } Catch {
-        Throw ("Get-NetView: Failed to create directory '$OutDir' because " + $error[0])
+        Throw ("Get-NetView : Failed to create directory '$OutDir' because " + $error[0])
```
tmolenh: Space added to be consistent with other Cmdlets on exception thrown (e.g. Set-VM)

PR #105, opened 2017-06-08
Get-NetView: Display parameters in final output

tmolenh:
## Changes
1. Renamed function `Version` to `Metadata` to cover slightly expanded scope.
2. Output the exact value of the parameters that were passed to Get-NetView in Metadata.txt
3. Use splatting to pass parameters to Get-NetView and Worker.

## Output
It will look like this:
`Parameters: [SkipAdminCheck, True] [OutputDirectory, .]`
If a parameter was not specified then it won't show up:
`Parameters: [SkipAdminCheck, True]`
If none are specified then it just looks like this:
`Parameter: `

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -1554,20 +1554,21 @@ function HwErrorReport {
     }
 } # HwErrorReport()
 
-function Version {
+function Metadata {
     [CmdletBinding()]
     Param(
-        [parameter(Mandatory=$true)] [String] $OutDir
+        [parameter(Mandatory=$true)] [String] $OutDir,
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -1722,7 +1723,7 @@ function Worker {
 
     EnvSetup          -OutDir $workDir
 
-    Version           -OutDir $workDir
+    Metadata          -OutDir $workDir -Params $PSBoundParameters.GetEnumerator()
```
tmolenh: It isn't possible to pass `$PSBoundParameters` directly because it is an automatic variable. AFAIK this is the best way to do this.

PR #110, opened 2017-06-28
Output to the temp folder if Desktop is undefined 

tmolenh:
On ServerCore, the desktop folder is not defined, so `GetFolderPath("Desktop")` will return an empty string. In this case, output to `$env:temp` instead.

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -1673,14 +1673,16 @@ function GetWorkDir {
     )
 
     # Validate user input or use desktop if none
-    $baseDir = if ($OutputDirectory) {
-                   if (Test-Path $OutputDirectory) {
+    $baseDir = If ($OutputDirectory) {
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -1673,14 +1673,16 @@ function GetWorkDir {
...
+    $baseDir = If ($OutputDirectory) {
+                   If (Test-Path $OutputDirectory) {
                        (Resolve-Path $OutputDirectory).Path # full path
-                   } else {
+                   } Else {
                        Throw "Get-NetView : The directory '$OutputDirectory' does not exist."
                    }
-               } else { # $OutputDirectory is undefined
-                   [Environment]::GetFolderPath("Desktop") + "\"
+               } ElseIf (($desktop = [Environment]::GetFolderPath("Desktop"))) {
```
tmolenh: Explanation: https://stackoverflow.com/a/7089056

PR #123, opened 2017-08-12
Update VMMExpress.ps1

CharbelNemnom:
The Storage Classification should be equal to "" otherwise the script is failing.


CharbelNemnom:
Thank you Greg!

PR #130, opened 2017-10-04
Adds Kubernetes On-Prem guide.

gkudra-msft:
This is intended to be a _temporary_ home for the tutorial until we get the kinks worked out. At that point, it can be put on the appropriate `MicrosoftDocs` repository.

I recommend pressing the "View" button on the Markdown file (`HOWTO-on-prem.md`) to see it [rendered properly](https://github.com/Microsoft/SDN/blob/df603505ca9a232ccf8124bdb39049b1304d2522/Kubernetes/HOWTO-on-prem.md).


Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+First, install all of the pre-requisites on the master:
+
+    # apt-get install git build-essentials docker.io
+
+### Building Kubernetes ###
+We will need to build the `kubelet` and `kubeproxy` binaries for Windows from scratch _cross-compiling from Linux_, so let's [set up a Go environment](https://golang.org/doc/install#install) for that, first. This guide assumes Go 1.9 and has not been tested with anything else. With your Go environment ready (don't forget to set your `$GOPATH`!), we can build the Windows binaries:
+
+```bash
+$ K8SREPO="github.com/madhanrm/kubernetes"
+$ go get -d $K8SREPO
```
JMesser81: Ignore error from this command, e.g. 'no Go files in $GOPATH/GitHub.com/madhanrm/kubernetes

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+$ cd $GOPATH/src/$K8SREPO
+$ KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubectl
```
JMesser81: Can you put this on two lines; it's confusing
$ KUBE_BUILD_PLATFORM=windows/amd64
$ make WHAT=cmd/kubectl

nwoodmsft: This syntax is pretty common on Linux when you only want the env variable to apply to the command being executed (in this case 'make'). If this is split across two lines, then the value of the env variable will persist for all future commands in that shell too. That might be ok, but just mentioning the side effect ;)

gkudra-msft: Sure, I can clarify that in the instructions and just unset the variable at the end.

JMesser81: Should be KUBE_BUILD_PLATFORMS (with an S)

JMesser81: If there are two lines for KUBE_BUILD_PLATFORMS and make command, make sure to export KUBE_BUILD_PLATFORMS environment variable
export KUBE_BUILD_PLATFORMS=windows/amd64

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+$ cp ~/_output/bin/kubectl.exe ~/kube/
```
JMesser81: Output is actually here: $GOPATH/src/GitHub.com/madhanrm/Kubernetes/_output/bin/

gkudra-msft: Good catch; should just be `_output/bin` then.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Building the Binaries ##
+We will begin with a Linux master node and a Windows worker node on RS3+, both of which are assumed to be VMs.
+
+First, install all of the pre-requisites on the master:
+
+    # apt-get install git build-essentials docker.io
+
+### Building Kubernetes ###
+We will need to build the `kubelet` and `kubeproxy` binaries for Windows from scratch _cross-compiling from Linux_, so let's [set up a Go environment](https://golang.org/doc/install#install) for that, first. This guide assumes Go 1.9 and has not been tested with anything else. With your Go environment ready (don't forget to set your `$GOPATH`!), we can build the Windows binaries:
```
JMesser81: If GOPATH is not set, packages may be brought into $GOROOT/pkg directory

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+$ go get -d $K8SREPO
+$ cd $GOPATH/src/$K8SREPO
+$ KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubectl
+$ cp ~/_output/bin/kubectl.exe ~/kube/
+$ git checkout cniwindows
+$ make clean && KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubelet
+$ cp ~/_output/bin/kubelet.exe ~/kube/
+$ git checkout winkernelproxy 
+$ KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kube-proxy
+$ cp ~/_output/bin/kube-proxy.exe ~/kube/
```
JMesser81: No '~/' in front of _output/ since these components aren't being built in user's home directory. Change to
cp _output/bin/kube-proxy.exe ~/kube/

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+_You may run into problems with various "permission denied" errors when building these binaries. This is a bug in the Kubernetes build process for Windows, which uses binaries that exist, but are size-zero, and thus don't have executable permissions on them. The solution, unfortunately, is to run `chmod +x _output/bin/*` after every time (~4 times) you hit the error._
+
+Done! Now, we also need the actual Linux Kubernetes binaries for v1.8. You can pull these from [the Kubernetes mainline](https://github.com/kubernetes/kubernetes/releases/tag/v1.8.0-rc.1), or build them from source as above, except using `K8SREPO=k8s.io/kubernetes` and the `release-1.8` branch. Copy these to `~/kube/bin` (note the extra `/bin` compared to the above path; we actual need to use these during master configuration, rather than just copying them to the Windows node).
```
JMesser81: Might want to add a note about downloading the tarball, running cluster/get-kube-binaries.sh

JMesser81: and tar xzvf server/Kubernetes-server-Linux-amd64.tar.gz

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Install CNI ###
```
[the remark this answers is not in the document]
gkudra-msft: Nope, we need the plugins. The Linux master uses the default `kubenet` plugin which needs those _plugin_ binaries.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Oh, I never actually tried it that way so I didn't know you ran a script to get the binaries.

JMesser81: Specifically, what server binaries are needed? hypercube or kube-apiserver or *.tar
Will we be running the K8s binaries in a container (need to import *.tar file) or directly on the host

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+We also need to install the basic CNI binaries so that networking works. Download them from [here](https://github.com/containernetworking/plugins/releases) and copy them to `/opt/cni/bin/`.
+
+## Prepare the Master ##
+Copy the entire directory from [this repository](https://github.com/Microsoft/SDN/tree/master/Kubernetes/linux), which contains a lot of the setup scripts and Kubernetes files that are essential to this process. Check them all out to `~/kube/` and make the scripts executable. This entire directory will be getting mounted for a lot of the docker containers in future steps, so keep its structure the same as outlined in the guide. 
+
+We'll be creating the service cluster on the virtual subnet 11.0/16. This isn't configurable, so if you want to tweak this, it'll require manual modification.
+
+### Certificates ###
+First, prepare the certificates that will be used for nodes to communicate inthe cluster. Run `generate-certs.sh` (in the `certs/` directory), passing **your** external IP as its (only) parameter -- this is likely whatever the IP address of your `eth0` interface, or whatever:
```
JMesser81: small nit: should mkdir certs/ and then mv/cp generate-certs.sh into certs/ directory

$ip addr show dev eth0 # to get ip address of eth0

Kubernetes/HOWTO-on-prem.md
gkudra-msft: `kubectl`, `hyperkube`, and `kubeadm` are required.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Heh, figured most Linux users would know the latter (or just `ifconfig`). But yes, good point regarding `certs/`. I changed the structure a bit throughout.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
+# Kubernetes On-Premises Deployment From Scratch -- Start to Finish #
+This guide will walk you through deploying *Kubernetes 1.8* on a Linux master and join two Windows nodes to it without a cloud provider.
+
+**Note**: The guide will assume a local working directory of `~/kube` for the Kubernetes setup. If you choose a different directory, just replace any references to that path.
+
+## Building the Binaries ##
+We will begin with a Linux master node and a Windows worker node on RS3+, both of which are assumed to be VMs.
```
nwoodmsft: RS3+ is the Windows node requirement, but what are the Linux master node requirements? Are we making any assumption on which distro of Linux the customer should be using to follow along with this guide?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+This guide will walk you through deploying *Kubernetes 1.8* on a Linux master and join two Windows nodes to it without a cloud provider.
+
+**Note**: The guide will assume a local working directory of `~/kube` for the Kubernetes setup. If you choose a different directory, just replace any references to that path.
+
+## Building the Binaries ##
+We will begin with a Linux master node and a Windows worker node on RS3+, both of which are assumed to be VMs.
+
+First, install all of the pre-requisites on the master:
+
+    # apt-get install git build-essentials docker.io
```
nwoodmsft: This requires root, so perhaps we should add 'sudo' to the front of this command for clarity. Alternatively, if there is an assumption that we will be elevated to root during all of this guide, then we could just have a pre-requisite step to execute 'sudo -s' or similar to get them to a root shell.

nwoodmsft: I think this is a typo, it should be "build-essential" without the s ? ("build-essentials" gave me an error)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+$ cp ~/_output/bin/kubectl.exe ~/kube/
+$ git checkout cniwindows
+$ make clean && KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubelet
+$ cp ~/_output/bin/kubelet.exe ~/kube/
+$ git checkout winkernelproxy 
+$ KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kube-proxy
+$ cp ~/_output/bin/kube-proxy.exe ~/kube/
+```
+
+_You may run into problems with various "permission denied" errors when building these binaries. This is a bug in the Kubernetes build process for Windows, which uses binaries that exist, but are size-zero, and thus don't have executable permissions on them. The solution, unfortunately, is to run `chmod +x _output/bin/*` after every time (~4 times) you hit the error._
```
nwoodmsft: "You may run into problems" - I did :) I think this is going to happen for most people, so this warning/guidance should come before the build steps so that the reader is already expecting the make commands to fail and knows what to do when they do fail.

nwoodmsft: Also, don't ask me why...but my binary ended up being output to the following location:
_output/local/bin/windows/amd64/kubectl.exe

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+```bash
+$ K8SREPO="github.com/madhanrm/kubernetes"
+$ go get -d $K8SREPO
+$ cd $GOPATH/src/$K8SREPO
+$ KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubectl
+$ cp ~/_output/bin/kubectl.exe ~/kube/
+$ git checkout cniwindows
+$ make clean && KUBE_BUILD_PLATFORM=windows/amd64 make WHAT=cmd/kubelet
+$ cp ~/_output/bin/kubelet.exe ~/kube/
```
nwoodmsft: There seems to be a missing step here. In my environment, I had to execute the 'make' for kubelet twice (once for Linux and once for Windows). If I don't first build for Linux, then commands like "defaulter-gen" won't be present in _output/bin and the Windows 'make' will fail. This seems to be what Madhan was indicating in his OneNote steps...it should be like this:

```
make clean && make WHAT=cmd/kubelet
KUBE_BUILD_PLATFORMS=windows/amd64 make WHAT=cmd/kubelet
```

nwoodmsft: Also, just like Madhan, my output binary was in: _output/local/bin/windows/amd64/kubelet.exe

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Prepare the Master ##
+Copy the entire directory from [this repository](https://github.com/Microsoft/SDN/tree/master/Kubernetes/linux), which contains a lot of the setup scripts and Kubernetes files that are essential to this process. Check them all out to `~/kube/` and make the scripts executable. This entire directory will be getting mounted for a lot of the docker containers in future steps, so keep its structure the same as outlined in the guide. 
+
+We'll be creating the service cluster on the virtual subnet 11.0/16. This isn't configurable, so if you want to tweak this, it'll require manual modification.
+
+### Certificates ###
+First, prepare the certificates that will be used for nodes to communicate inthe cluster. Run `generate-certs.sh` (in the `certs/` directory), passing **your** external IP as its (only) parameter -- this is likely whatever the IP address of your `eth0` interface, or whatever:
+
+    ~/kube/certs $ ./generate-certs.sh 10.123.45.67   # example! replace
```
nwoodmsft: Just a thought...since the master IP needs to be passed to several commands, why don't we just have them assign it to an environment variable and then refer to that variable for the rest of the guide?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Allow Routing ###
+_This step may be optional, depending on whether or not you've configured your intended cluster subnet to be routable already_.
+
+Windows nodes will each grab a /24 from the cluster, so if your cluster CIDR is 192.168.0.0/16, a Windows node will use 192.168.0.1/24. Just run:
+
+    ~/kube $ generate-routes.sh 192.168
```
nwoodmsft: I pulled down these scripts to my master node by cloning https://github.com/Microsoft/SDN and switching to the k8s-guide branch. The generate-certs.sh script worked fine, but generate-routes.sh and start-kubelet.sh both failed to run properly and it seemed to be related to Windows line endings making their way into those files. I recreated those files manually (copy/pasting the raw content into the file on linux) and then they worked...so plz double check the scripts on GitHub just in case there is something fishy :)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+    ~/kube $ ./configure-kubectl.sh 10.123.45.67
+
+Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, copy the Kubernetes configuration file to the local kubelet mount path:
+
+    $ sudo cp ~/.kube/config kubelet/
+
+**TODO**: Why is this necessary? Shouldn't this be done automatically by one of the containers?
+
+In another terminal session, run the Kubeproxy script, passing your cluster CIDR (you may also need to run this one with `sudo`):
```
nwoodmsft: I think the script name is "start-kubeproxy.sh"

Kubernetes/HOWTO-on-prem.md
nwoodmsft: Same problem with start-kubeproxy.sh script too:

> root@nick-linux3:~/kube# ./start-kubeproxy.sh 192.160
> ./start-kubeproxy.sh: line 6: $'\r': command not found

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+    ~/kube $ ./kubeproxy.sh 192.168
+
+After some time, you should see the following system state:
+
+  - Under `docker ps`, you should see worker and pod containers mirroring the configs in `manifest/` and in `addons/`.
+  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master, DNS, and heapster.
+  - A new interface `cbr0` under `ipconfig`, with your chosen cluster CIDR.
```
nwoodmsft: ifconfig

Kubernetes/linux/addons/kube-dns-depolyment.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: v1
```
nwoodmsft: Filename has a typo "depolyment"

Kubernetes/linux/manifest/generate.py
```diff
@@ -0,0 +1,260 @@
...
+    - name: addons
+      mountPath: "/etc/kubernetes/addons"
+      readOnly: true
+  volumes:
+  - name: addons
+    hostPath:
+      path: "$HOME/kube/addons"
+""",
+    "kube-apiserver.yaml": """
+    apiVersion: "v1"
```
nwoodmsft: The whitespace here before "apiVersion" is a problem. It seemed to cause the resultant kube-apiserver.yaml in my deployment to be invalid and therefore apiserver failed to come up as a result. I manually removed the whitespace from the yaml and then things started working.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+    $ sudo cp ~/.kube/config kubelet/
+
+**TODO**: Why is this necessary? Shouldn't this be done automatically by one of the containers?
+
+In another terminal session, run the Kubeproxy script, passing your cluster CIDR (you may also need to run this one with `sudo`):
+
+    ~/kube $ ./kubeproxy.sh 192.168
+
+After some time, you should see the following system state:
```
nwoodmsft: "After some time" - people get impatient quickly (I do!) so do we know a rough timeframe? Are we talking 2 minutes or should folks wait ~15 minutes before checking for the system state you mention.

Kubernetes/linux/manifest/generate.py
```diff
@@ -0,0 +1,260 @@
...
+        path: "$HOME/kube"
+    - name: "var-lib-kubelet"
+      hostPath:
+        path: "$HOME/kube/kubelet"
+    - name: "varlog"
+      hostPath:
+        path: "$HOME/kube/log/kube-controller-manager"
+""",
+    "kube-scheduler.yaml": """
+    apiVersion: "v1"
```
nwoodmsft: Same problem here, the whitespace before apiVersion for kube-scheduler.yaml is causing problems

Kubernetes/HOWTO-on-prem.md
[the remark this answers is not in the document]
JMesser81: Got it. Can you make this section more clear?

Install CNI *Plugins*
We also need to install the basic CNI *plugin* binaries so that networking works. Download *the plugins*

Kubernetes/HOWTO-on-prem.md
JMesser81: I think people following these instructions may be Windows users and not have much experience with Linux.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Allow Routing ###
+_This step may be optional, depending on whether or not you've configured your intended cluster subnet to be routable already_.
+
+Windows nodes will each grab a /24 from the cluster, so if your cluster CIDR is 192.168.0.0/16, a Windows node will use 192.168.0.1/24. Just run:
+
+    ~/kube $ generate-routes.sh 192.168
+
+### Prepare Manifests & Addons ###
+In the `manifest` folder, run the Python script, passing your master IP and the _full_ cluster CIDR:
```
JMesser81: In my environment, the cluster CIDR (i.e. the IP subnet of the cluster node/host IPs) is 10.127.132.128/25 while the POD CIDR ranges are different per host - 10.10.185.0/26, 10.10.185.64/26, and 10.10.185.128/26. Should the "full" cluster CIDR be 10.0.0.0/8? What if my POD CIDR range was in 192.168.x.y/24 ?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+This will generate a set of YAML files. You should [re]move the Python script so that K8s doesn't mistake it for a manifest and cause problems.
+
+### Configure & Run Kubernetes ###
+Run the script `configure-kubectl.sh`, passing your external IP again, to create a configuration in `~/.kube/config` containing all of the certificate data and other parameters:
+
+    ~/kube $ ./configure-kubectl.sh 10.123.45.67
+
+Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, copy the Kubernetes configuration file to the local kubelet mount path:
+
+    $ sudo cp ~/.kube/config kubelet/
```
JMesser81: I receive multiple errors upon running this script including Windows CR-LF (/r) character: command not found. Looks like script must be run with sudo for access various directories such as /var/run/Kubernetes/

JMesser81: I assume we need to create kubelet/ directory first? Or, is this "local kubelet mount path" somewhere in /etc or /var?

Kubernetes/HOWTO-on-prem.md
JMesser81: The error I keep seeing is

E1009 09:15:05.072900   94524 reflector.go:205] k8s.io/kubernetes/pkg/client/informers/informers_generated/internalversion/factory.go:73: Failed to list *api.Service: Get https://10.127.132.211/api/v1/services?resourceVersion=0: dial tcp 10.127.132.211:443: getsockopt: connection refused

is this expected?

nwoodmsft: If ApiServer is not running, then it might be due to the issue I mentioned in generate.py (whitespace before "apiVersion" in the files generated by generate.py. That was causing my ApiServer to not come up.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+    ~/kube $ ./kubeproxy.sh 192.168
+
+After some time, you should see the following system state:
+
+  - Under `docker ps`, you should see worker and pod containers mirroring the configs in `manifest/` and in `addons/`.
+  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master, DNS, and heapster.
+  - A new interface `cbr0` under `ipconfig`, with your chosen cluster CIDR.
+
+## Join a Windows Node ##
+Like for Linux, we need an assortment of scripts to prepare things for us. They can be found [here](https://github.com/Microsoft/SDN/tree/master/Kubernetes/windows). Place these in a new folder, `C:\k\`. We also need to copy the .exe binaries we compiled earlier (which we placed in `~/kube/`), as well as the Kubectl config file (from `~/.kube/config`), into this folder.
```
nwoodmsft: If this guide is aimed at Windows users with only basic Linux exposure, then we might need to explain how they get the compiled binaries out of the Linux VM and into the Windows VM?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+to `C:\k\` and run:
+
+    C:\k> docker pull microsoft/windowsservercore-insider:latest
+    C:\k> docker tag [SHA from previous cmd] microsoft/windowsservercore:latest
+    C:\k> docker build -t kublet/pause .
+
+### Join to Cluster ###
+In two PowerShell windows, simply run:
+
+    PS> ./start-kubelet.ps1 -ClusterCidr [Cluster CIDR]
```
nwoodmsft: Elsewhere in the guide we differentiate between passing full CIDR or just the first two octets...maybe we should be explicit here too

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,103 @@
...
+## Join a Windows Node ##
+Like for Linux, we need an assortment of scripts to prepare things for us. They can be found [here](https://github.com/Microsoft/SDN/tree/master/Kubernetes/windows). Place these in a new folder, `C:\k\`. We also need to copy the .exe binaries we compiled earlier (which we placed in `~/kube/`), as well as the Kubectl config file (from `~/.kube/config`), into this folder.
+
+### Build Docker Image ###
+We need to build the docker image for the Kubernetes infrastructure. Navigate
+to `C:\k\` and run:
+
+    C:\k> docker pull microsoft/windowsservercore-insider:latest
+    C:\k> docker tag [SHA from previous cmd] microsoft/windowsservercore:latest
+    C:\k> docker build -t kublet/pause .
```
nwoodmsft: Is "kublet" a typo?

nwoodmsft: Either way, the scripts (start-kubelet.ps1) refer to this image name instead: kubletwin/pause

gkudra-msft: Ah, oops. The `kublet` part is intentional -- a leftover from Madhan's typo. :)

Kubernetes/HOWTO-on-prem.md
gkudra-msft: The pod CIDR is supposed to be a /16, with each node getting a /24 for its pods out of that. I will clarify that in the guide.

Kubernetes/linux/manifest/generate.py
gkudra-msft: Dang, I thought I fixed that already before committing. Thanks!

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Should be about a minute -- will clarify.

Kubernetes/linux/addons/kube-dns-depolyment.yaml
gkudra-msft: Yeah, I had fixed that at some point, then reverted back because I was scrambling for things to blame for not working. :)

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Ah, that's interesting. This may mean people won't need to fiddle with the italicized "if you get permission errors" step. I'll try this out to confirm and edit the guide to reflect that. Good tip!

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Good point -- I'll make note of it to just say "a recent Ubuntu-like distro".

Kubernetes/HOWTO-on-prem.md
gkudra-msft: The `#` was a hint at that (vs. the standard `$`), but I will clarify since it's probably too subtle, especially for those unfamiliar with Linux!

Kubernetes/HOWTO-on-prem.md
gkudra-msft: It's important to differentiate the CIDRs here. "Cluster CIDR" is the subnet for _pods across nodes_. What you describe as your cluster CIDR (`10.127.132.128/25`) is simply the external network that k8s works across and is only relevant in the few steps to -- for example -- expose the API server to be accessible from the interface bound to the external network.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: This is created by one of the master containers (full path would be `~/kube/kubelet`).

Kubernetes/HOWTO-on-prem.md
JMesser81: +1

Kubernetes/HOWTO-on-prem.md
```diff
@@ -1,53 +1,72 @@
...
+    ~/kube/certs $ mv ../generate-certs.sh .
     ~/kube/certs $ ./generate-certs.sh 10.123.45.67   # example! replace
 
 ### Allow Routing ###
 _This step may be optional, depending on whether or not you've configured your intended cluster subnet to be routable already_.
 
-Windows nodes will each grab a /24 from the cluster, so if your cluster CIDR is 192.168.0.0/16, a Windows node will use 192.168.0.1/24. Just run:
+Windows nodes will each grab a /24 from the cluster CIDR, so it's 192.168.0.0/16, the first Windows node will use 192.168.0.1/24 for its pods. Pass the first two octets of your clister CIDR to generate the routes:
 
     ~/kube $ generate-routes.sh 192.168
```
JMesser81: This assumes that we have a contiguous "/16" available which is usually not the case in an L3-routed network topology. Is there a way to shorten this POD CIDR? For instance, I only have three /26s which I need to give explicitly to my three nodes. They are contiguous but they don't start at 0 (e.g. 10.10.185.0/26, 10.10.185.64/26, 10.10.185.128/26)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -59,41 +78,41 @@ In the `manifest` folder, run the Python script, passing your master IP and the
...
+Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, watch your working directory (`~/kube`) until the folder `kubelet/` appears. Immediately copy the Kubernetes configuration file to it:
 
-    $ sudo cp ~/.kube/config kubelet/
+    ~/kube $ sudo cp ~/.kube/config kubelet/
 
 **TODO**: Why is this necessary? Shouldn't this be done automatically by one of the containers?
 
 In another terminal session, run the Kubeproxy script, passing your cluster CIDR (you may also need to run this one with `sudo`):
 
     ~/kube $ ./kubeproxy.sh 192.168
```
JMesser81: I get an error from kube-proxy complaining about conntrack not being available.
E1010 15:55:19.407842  103781 proxier.go:942] Failed to delete kube-system/kube-dns:dns endpoint connections, error: error deleting conntrack entries for UDP peer {11.0.0.10, 10.10.0.9}, error: error looking for path of conntrack: exec: "conntrack": executable file not found in $PATH

Kubernetes/HOWTO-on-prem.md
```diff
@@ -59,41 +78,41 @@ In the `manifest` folder, run the Python script, passing your master IP and the
...
 
-After some time, you should see the following system state:
+After a few minutes, you should see the following system state:
 
   - Under `docker ps`, you should see worker and pod containers mirroring the configs in `manifest/` and in `addons/`.
-  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master, DNS, and heapster.
-  - A new interface `cbr0` under `ipconfig`, with your chosen cluster CIDR.
+  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master API server, plus the DNS and heapster addons.
+  - A new interface `cbr0` under `ifconfig`, with your chosen cluster CIDR.
```
JMesser81: I see cbr0 but per my earlier comment, I need the ability to specify more explicitly what the POD CIDR should be. As the one currently assigned is incorrect

Kubernetes/HOWTO-on-prem.md
```diff
@@ -59,41 +78,41 @@ In the `manifest` folder, run the Python script, passing your master IP and the
 This will generate a set of YAML files. You should [re]move the Python script so that K8s doesn't mistake it for a manifest and cause problems.
 
 ### Configure & Run Kubernetes ###
-Run the script `configure-kubectl.sh`, passing your external IP again, to create a configuration in `~/.kube/config` containing all of the certificate data and other parameters:
+Run the script `configure-kubectl.sh` passing your external IP again. This will create a configuration in `~/.kube/config` containing all of the certificate data and other parameters:
 
     ~/kube $ ./configure-kubectl.sh 10.123.45.67
 
-Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, copy the Kubernetes configuration file to the local kubelet mount path:
+Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, watch your working directory (`~/kube`) until the folder `kubelet/` appears. Immediately copy the Kubernetes configuration file to it:
```
JMesser81: I'm seeing multiple errors with kube-dns pods. Multiple CrashLoopBackOff and restarts

Kubernetes/HOWTO-on-prem.md
```diff
@@ -59,41 +78,41 @@ In the `manifest` folder, run the Python script, passing your master IP and the
...
-After some time, you should see the following system state:
+After a few minutes, you should see the following system state:
 
   - Under `docker ps`, you should see worker and pod containers mirroring the configs in `manifest/` and in `addons/`.
-  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master, DNS, and heapster.
-  - A new interface `cbr0` under `ipconfig`, with your chosen cluster CIDR.
+  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master API server, plus the DNS and heapster addons.
+  - A new interface `cbr0` under `ifconfig`, with your chosen cluster CIDR.
 
 ## Join a Windows Node ##
```
JMesser81: TODO: We should add instructions for adding a Linux node as well (or point to link in K8s documentation)

Kubernetes/HOWTO-on-prem.md
nwoodmsft: I'm not sure how conntrack is being used, but I also saw this error. You can resolve it by apt-get'ing conntrack, so we should probably add this to the guide along with the step where we install docker etc.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,122 @@
...
+
+    ~/kube $ ./kubeproxy.sh 192.168
+
+After a few minutes, you should see the following system state:
+
+  - Under `docker ps`, you should see worker and pod containers mirroring the configs in `manifest/` and in `addons/`.
+  - A call to `~/kube/bin/kubectl cluster-info` should show info for the Kubernetes master API server, plus the DNS and heapster addons.
+  - A new interface `cbr0` under `ifconfig`, with your chosen cluster CIDR.
+
+## Join a Windows Node ##
```
nwoodmsft: I just realized that we didn't tell customers what is needed on the Windows node :) For example, this is assuming that they have docker...are there any other assumptions around the version/sku of Windows or other pre-reqs? Can we provide instructions for getting docker installed?

rjmorse:
I've targeted this branch with a PR in https://github.com/Microsoft/SDN/pull/132 with my feedback since learning that this was online yesterday. It addresses some of the comments previously discussed as well as items I've faced in testing.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+
+```bash
+DIST_DIR="${HOME}/kube/"
+mkdir ${DIST_DIR}
+SRC_DIR="${HOME}/src/kubernetes/"
+mkdir -p "${SRC_DIR}"
+git clone https://github.com/madhanrm/kubernetes.git ${SRC_DIR}
+cd ${SRC_DIR}
+git checkout cniwindows
+build/run.sh make WHAT=cmd/kubelet KUBE_BUILD_PLATFORMS=linux/amd64
```
nwoodmsft: Following these instructions (from scratch and exactly) I had an issue with this build/run.sh command because docker was installed by apt but the service wasn't started. Then after starting the service with 'sudo service docker start' the command still failed because I was not root or using sudo + my user was not in the 'docker' group.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+build/run.sh make WHAT=cmd/kubelet KUBE_BUILD_PLATFORMS=windows/amd64
+cp ${SRC_DIR}/_output/dockerized/bin/windows/amd64/kubelet.exe ${DIST_DIR}
+#winkernelproxy has already been merged but still needs to be built
+git checkout winkernelproxy 
+build/run.sh make WHAT=cmd/kube-proxy KUBE_BUILD_PLATFORMS=windows/amd64
```
nwoodmsft: Before this run.sh command, you need to execute it for linux/amd64 to build kube-proxy, otherwise windows build will fail.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+cp ${SRC_DIR}/_output/dockerized/bin/windows/amd64/kube-proxy.exe ${DIST_DIR}
+ls ${DIST_DIR}
+```
+
+Done! Now, we also need the actual Linux Kubernetes binaries for v1.8. You can pull these from [the Kubernetes mainline](https://github.com/kubernetes/kubernetes/releases/tag/v1.8.0), or build them from source as above, except using `K8SREPO=k8s.io/kubernetes` and the `release-1.8` branch, stopping before the `export ...` line.
```
nwoodmsft: What 'export ..." line?

nwoodmsft: Maybe this block needs to be written now build steps have been tweaked. I don't know what K8REPO= references now.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: I haven't actually tried the dockerized steps, so thanks. I'll link to [this answer](https://askubuntu.com/a/477554).

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Yeah this was badly merged by me. I'm fixing it. :)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+
+If you built them from source, copy the binaries (at least `hyperkube`, `kubectl`, and `kubeadm`) directly to `~/kube/bin`. Otherwise, you will need to extract the downloaded archive and run the `cluster/get-kube-binaries.sh` script before copying the binaries.
+
+Copy these to `~/kube/bin` (note the extra `/bin` compared to the above path; we actual need to use these during master configuration, rather than just copying them to the Windows node). The v1.8 Windows binary `kubectl.exe` is included in the `windows/` directory of this repo, but you can also build it here (setting `KUBE_BUILD_PLATFORMS=windows/amd64` and `WHAT=cmd/kubectl` as above).
```
nwoodmsft: Obtaining kubectl.exe became more confusing now...we should just have them build kubectl above along with the rest of the bins.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+### Install CNI Plugins ###
+
+We also need to install the basic CNI plugin binaries so that networking works. Download them from [here](https://github.com/containernetworking/plugins/releases) and copy them to `/opt/cni/bin/`.
+
+```bash
+DOWNLOAD_DIR="${HOME}/kube/cni-plugins"
+CNI_BIN="/opt/cni/bin/"
+mkdir ${DOWNLOAD_DIR}
+cd $DOWNLOAD_DIR
+curl -L $(curl -s https://api.github.com/repos/containernetworking/plugins/releases/latest | grep browser_download_url | grep 'amd64.*tgz' | head -n 1 | cut -d '"' -f 4) -o cni-plugins-amd64.tgz
```
nwoodmsft: Requires curl, do you want to ask them to apt install this earlier on ?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -0,0 +1,171 @@
...
+
+Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, watch your working directory (`~/kube`) until the folder `kubelet/` appears. Immediately copy the Kubernetes configuration file to it:
+
+    ~/kube $ sudo cp ~/.kube/config kubelet/
+
+**TODO**: Why is this necessary? Shouldn't this be done automatically by one of the containers?
+
+In another terminal session, run the Kubeproxy script, passing your cluster CIDR (you may also need to run this one with `sudo`):
+
+    ~/kube $ ./kubeproxy.sh 192.168
```
nwoodmsft: This filename is start-kubeproxy.sh

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Is curl really not included in almost every distro? :) added!

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Didn't see this till now. I'm cool with it, but there's the caveat of them not closing the shell. I'll be sure to clarify. :)

PR #133, opened 2017-10-13
Improve Get-VSwitchPathCost.ps1 performance

tmolenh:
Instead of collecting each counter in its own Job, collect them all at once because it much faster.

### DIFF
--- a/Diagnostics/Get-VSwitchPathCost.ps1
+++ b/Diagnostics/Get-VSwitchPathCost.ps1
@@ -30,8 +30,10 @@
     the measurement. This time period will be added to Duration for total runtime.
     Default value is 1 second.
 .PARAMETER Force
-    Switch parameter that forces the scipt to run while Hyper-Threading is enabled (don't prompt).
-    Does nothing if Hyper-Threading is disabled.
+    Switch parameter that suppresses continue prompts, and will default to 'yes'.
+    Without this switch, the user be prompted to continue if a power saving profile 
+    or Hyper-Threading is enabled. If none of these features are enabled, then this
+    Switch will do nothing.
 .EXAMPLE
     Get-VSwitchPathCost.ps1 -Duration 30 -SwitchName MyVSwitch
 .NOTES
@@ -40,7 +42,7 @@
     The script only collects perfmon counters for the specified during, and computes
     path cost, based on the specified CPU range.
 
-    Note that CPU utilization and path cost are most accurately measured with Hyper-Threading
+    Note that CPU utilization and path cost can only be accurately measured with Hyper-Threading
     disabled. You can disable Hyper-Threading in the BIOS. However, note that for better
     performance, it should be enabled.
 #>
@@ -78,12 +80,13 @@ function Log($Message)
 {
     if ($Message.StartsWith("ERROR"))
     {
-        Write-Error "$Message"
-        exit 1
+        $msg = ($Message -split "ERROR: ")[1]
+        throw "Get-VSwitchPathCost : $msg"
     }
     elseif ($Message.StartsWith("WARNING"))
     {
-        Write-Warning "$Message"
+        $msg = ($Message -split "WARNING: ")[1]
+        Write-Warning "$msg"
     }
     else
     {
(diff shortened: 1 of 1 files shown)


omarcardona:
I'll leave the Squash and Merge steps to Khoa.  Please make sure he reviews before final integration.

Diagnostics/Get-VSwitchPathCost.ps1
```diff
@@ -127,34 +136,25 @@ function Get-CounterAverage([String[]] $statArray)
 
 if (IsHyperThreadingEnabled)
 {
-    Log "ERROR: Hyper-Threading is enabled. Please disable Hyper-Threading before starting the test."
+    Log "WARNING: Hyper-Threading is enabled. This may affect the accuracy of the results."
 }
 
-$switch = Get-VMSwitch $SwitchName
```
tmolenh: This cmdlet is the only part of the script that requires admin, so I removed it in favor of checking if the counters exist. However, I don't check for duplicate switch names now.

Diagnostics/Get-VSwitchPathCost.ps1
```diff
@@ -127,34 +136,25 @@ function Get-CounterAverage([String[]] $statArray)
...
-    $MaxCpuNumber = $cpuPerNuma * $numaCount - 1
+    $MaxCpuNumber = Get-MaxCpuNumber
     Log "Set MaxCpuNumber to $MaxCpuNumber"
 }
-
-if ($MaxCpuNumber -lt $BaseCpuNumber)
+elseif (IsHyperThreadingEnabled)
 {
-   Log "ERROR: Max CPU number is less than base CPU number"
+    # force MaxCpuNumber to odd value with HT so the entire core is measured
```
tmolenh: This is correct, right?

Diagnostics/Get-VSwitchPathCost.ps1
```diff
@@ -1,74 +1,52 @@
...
    This script does not start any traffic. It is the caller's responsibility to ensure
    traffic is already running in steady state, before calling the script.
    The script only collects perfmon counters for the specified during, and computes
    path cost, based on the specified CPU range.
 
-   Since CPU utilization and path cost are more accurately measured with HyperThreading
-   disabled, this script forces HyperThreading to be disabled. You can disable HyperThreading
-   in the BIOS.
-
    However, note that for better performance, HyperThreading should be enabled.
```
mskhot: should remove this line, if you remove the paragraph above?

Diagnostics/Get-VSwitchPathCost.ps1
```diff
@@ -88,98 +66,82 @@ function Log($Message)
     }
 }
 
+$cpuInfo = @(Get-WmiObject -Class win32_processor -Property NumberOfCores, MaxClockSpeed, NumberOfLogicalProcessors)
```
mskhot: Can you check to see if there's a property for currentClockSpeed, and throw an error if MaxClockSpeed != CurrentClockSpeed?  In the BIOS, by default, the machine uses a PowerSaving profile, and the CPU speed is much lower than current clock speed.

We want the power profile to be "Performance", when doing perf measurements, so the CPU speed doesn't vary.

Diagnostics/Get-VSwitchPathCost.ps1
mskhot: No, with HyperThreading enabled, we can't really deterministically measure the CPU cycles. 
Also, the maxCpuNumber calculation below is not correct.  This means you're ignoring half of the CPUs.  The HT core pairs are 0/1, 2/3, 4/5, etc.

Diagnostics/Get-VSwitchPathCost.ps1
```diff
@@ -1,293 +1,254 @@
...
-}
-
-#
-# Verify HyperThreading is disabled
-#
-if (IsHyperThreadingEnabled)
+elseif (IsHyperThreadingEnabled)
 {
-    Log "ERROR: HyperThreading is enabled. Please disable HyperThreading before starting the test."
+    # force MaxCpuNumber to odd value with HT so the entire core is measured
```
mskhot: maybe print a warning?

tmolenh: Just realized that BaseCpuNumber should be forced to even value as well.

PR #139, opened 2017-11-02
Updates Kubernetes guide.

gkudra-msft:
In detail, I have:

 - improved the build instructions to show a consistent version across the binaries and removed several steps to build binaries from scratch in favor of pre-built ones;

 - added some sections to expand on custom pod CIDRs, for when infrastructure doesn't support a full /16 subnet;

 - added more "copy-pastable" sections in favor of text;

 - cleaned up a lot of the language and the command-lines, and improved section ordering somewhat (imo).

- made an important distinction regarding the Linux master, which is that the manual setup isn't technically necessary; if someone has an existing cluster, they just need to copy the certificate configuration file (such as `/etc/kubernetes/admin.conf` if deployed through `kubeadm`) to the Windows node for it to work.

Also, tagging @madhanrm as an FYI because I can't add him as a reviewer.


gkudra-msft:
As it turns out, the CNI PR (kubernetes/kubernetes#51063) will not make it to the 1.8 release. We're merging it directly to the mainline. Thus, this should _probably_ be updated to refer to 1.9 in the various places, but I'm not sure. Kubernetes is supposed to be [`N-1`-client compatible](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#version-skew-policy):

> one minor version skew between the kubelet and the control plane is supported, but the kubelet version may never exceed the API server version. For example, kubelets running 1.7.0 should be fully compatible with a 1.8.0 API server.

Regardless, I'll defer to the reviewers to comment on that.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
+
+```bash
+$ KUBEREPO="k8s.io/kubernetes"
+$ go get -d $KUBEREPO
+# Note: the above command may spit out a message about 
+#       "no Go files in...", but it can be safely ignored!
+$ cd $GOPATH/src/$KUBEREPO
+# if you want the "bleeding edge" Kubernetes, run 
+# `git checkout release-1.8` here.
+$ git checkout tags/v1.8.2
```
JMesser81: Shouldn't this be tags/v1.9.0-alpha.2 which is the latest v1.9 alpha release?

gkudra-msft: [Yes,](https://github.com/Microsoft/SDN/pull/139#issuecomment-341799257) I was waiting on a confirmation. :)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
+$ make clean && make WHAT=cmd/kubelet
+
+# necessary to add a bug-fix not yet upstream'd
+$ git add remote patch https://github.com/madhanrm/kubernetes
+$ git fetch patch
+$ git cherry-pick cba7ee2a4ee64bd4aafafa403d583310a49853fd
```
JMesser81: Looks like most recent commit for kube-proxy fix (https://github.com/kubernetes/kubernetes/pull/54969) is here

723bfbaf1c40619e3a251bb7915b2ae313a2648f

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
+#       "no Go files in...", but it can be safely ignored!
+$ cd $GOPATH/src/$KUBEREPO
+# if you want the "bleeding edge" Kubernetes, run 
+# `git checkout release-1.8` here.
+$ git checkout tags/v1.8.2
+$ make clean && make WHAT=cmd/kubelet
+
+# necessary to add a bug-fix not yet upstream'd
+$ git add remote patch https://github.com/madhanrm/kubernetes
+$ git fetch patch
```
JMesser81: Is this patch specific to v1.8.2 or will it work with latest upstream (and latest alpha build?)

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
+# Note: the above command may spit out a message about 
+#       "no Go files in...", but it can be safely ignored!
+$ cd $GOPATH/src/$KUBEREPO
+# if you want the "bleeding edge" Kubernetes, run 
+# `git checkout release-1.8` here.
+$ git checkout tags/v1.8.2
+$ make clean && make WHAT=cmd/kubelet
+
+# necessary to add a bug-fix not yet upstream'd
+$ git add remote patch https://github.com/madhanrm/kubernetes
```
JMesser81: What's the difference between doing this and just grabbing upstream code (or v1.9.2 alpha) and pulling in PRs directly?

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Oh, nice. That PR has [4 commits](https://github.com/kubernetes/kubernetes/pull/54969/commits), so we'll have to apply them all. The one in the guide is _only_ the port fix.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: That's pretty much exactly what this is doing. `git` doesn't have a concept of a "PR," so we just grab the commits from the PR directly.

Kubernetes/HOWTO-on-prem.md
gkudra-msft: Once the PR you linked goes in, it won't be necessary. Otherwise, it's always needed.

Kubernetes/HOWTO-on-prem.md
JMesser81: Seems to work although I had to manually deal with some merge conflicts.

JMesser81: We shouldn't need to fetch the "patch" branch (Madhan's branch) if we're using v1.9.0-alpha.2 correct?

gkudra-msft: Shoot, that's annoying. Yes, if it's merged before the next release (Nov. 15th, apparently) the entire "Building Kubernetes' Windows Binaries" section can go.

gkudra-msft: Were the merge conflicts code-related or just build scripts etc.? Can we expect a user to be able to resolve them on their own?

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
 
-Windows nodes will each grab a /24 from the cluster CIDR, so it's 192.168.0.0/16, the first Windows node will use 192.168.0.1/24 for its pods. Pass the first two octets of your clister CIDR to generate the routes:
-
-    ~/kube $ generate-routes.sh 192.168
 
 ### Prepare Manifests & Addons ###
 In the `manifest` folder, run the Python script, passing your master IP and the _full_ cluster CIDR:
 
-    $ python2 generate.py 10.123.45.67 192.168.0.0/16
+    $ ./generate.py $MASTER_IP 192.168.0.0/16
```
JMesser81: We need an extra option on this script to specify topology / configuration being used:
1) Upstream Routed (Static Routes in ToR)
2) Host-Gateway (Static Routes on each Host)
3) (Optional) Flannel with win-overlay CNI plugin

This will not work if user is running in "Upstream Routed" mode where static routes are configured in the ToR. In this case, kube-control-manager.yaml file should not have "--allocate-node-cidrs=True" or "--cluster-cidr=10.10.0.0/16". Instead, each instance of kubelet running on a particular node will need to have an option "--cidr=X" to specify the static CIDR

JMesser81: Also, this version still specifies v1.7.3 for kube-apiserver instead of v1.9.0

gkudra-msft: Yep, absolutely true. I added [this section](https://github.com/Microsoft/SDN/blob/6d1b667aa617e3371a67af30e12dd5f7a309517d/Kubernetes/HOWTO-on-prem.md#optional-customize-cluster-cidr) to do that manually (though I forgot to remove `--cluster-cidr`). 

You're saying we should make it parameterized instead, yes?

gkudra-msft: Waiting on @nwoodmsft for the versioning. :)

JMesser81: Ah... didn't see the next section below. Can we integrate that option into the script?

gkudra-msft: Definitely. Since the script is getting a little more complicated, I'll take the time to make it a bit nicer.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
 
-    ~/kube $ ./configure-kubectl.sh 10.123.45.67
+Finally, we're ready to start `kubelet`, the Kubernetes "client." This script runs indefinitely, so open another shell afterward to keep working:
 
-Run the script `start-kubelet.sh`. You may need to run it with `sudo`. Then, in another shell, watch your working directory (`~/kube`) until the folder `kubelet/` appears. Immediately copy the Kubernetes configuration file to it:
+    ~/kube $ sudo ./start-kubelet.sh
+
+Watch your working directory until the `kubelet` _folder_ appears in it. Immediately copy the Kubernetes configuration file to it:
 
     ~/kube $ sudo cp ~/.kube/config kubelet/
```
JMesser81: I still don't get why we have to do this... :-)

gkudra-msft: I wish I knew :P

Kubernetes/HOWTO-on-prem.md
```diff
@@ -102,91 +75,219 @@ sudo cp -r !(*.tgz) ${CNI_BIN}
...
 
-**TODO**: Why is this necessary? Shouldn't this be done automatically by one of the containers?
+You may need to restart the `start-kubelet.sh` script after this if it wasn't picked up fast enough.
+
+> This is a quirky workaround because for whatever reason, the API server pod doesn't mount the configuration file like it should.
+
+In yet another terminal session, run the Kubeproxy script, passing your cluster CIDR
 
-In another terminal session, run the Kubeproxy script, passing your cluster CIDR (you may also need to run this one with `sudo`):
+    ~/kube $ sudo ./start-kubeproxy.sh 192.168
```
JMesser81: Does it matter if we're using upstream routed mode with static POD CIDRs specified on each host if the cluster CIDR referenced isn't a full /16? Can it be any prefix with mask that matches the POD CIDR range across the cluster?

JMesser81: Can we start both kubelet and kube-proxy referenced in these scripts (start-kubeproxy.sh and start-kubelet.sh) in the background using "&" and then redirect stdout and stderr to some log file?

gkudra-msft: Definitely. This is something users can just do on their own as they wish (they can even make them a systemd service that runs on startup, etc.) but I'll add a note.

gkudra-msft: As for Kubeproxy, that's actually a really good question. I'm not sure. The [documentation](https://kubernetes.io/docs/admin/kube-proxy/) says:

 > `--cluster-cidr string`
>
> The CIDR range of pods in the cluster. When configured, traffic sent to a Service cluster IP from outside this range will be masqueraded and traffic sent from pods to an external LoadBalancer IP will be directed to the respective cluster IP instead

This implies to me that if you have a weird cluster configuration, where, say, master is on 192.168.0.0/28 and node 1 is on 192.168.5.0/31, you would *still* have to specify 192.168/16. **But** because it only applies to traffic going to service IPs and load balancer IPs, it wouldn't interfere with traffic on other subnets that isn't destined for these. 

Really good catch; I'll make note of that in this section.

Kubernetes/HOWTO-on-prem.md
```diff
@@ -1,94 +1,67 @@
...
+We can pull most of the binaries we need from the official Kubernetes repository, but will need to cross-compile one of the Windows binaries ourselves because it isn't "officially released."
 
-**Note**: If you run into "permission denied" errors, these can be avoided by building the Linux kubelet first, per the note in [acs-engine](https://github.com/Azure/acs-engine/blob/master/scripts/build-windows-k8s.sh#L176):
+There is a collection of scripts in [this repository](https://github.com/Microsoft/SDN/tree/master/Kubernetes/linux), which will help us with the setup process. Check them all out to `~/kube/` and make the scripts executable. This entire directory will be getting mounted for a lot of the docker containers in future steps, so keep its structure the same as outlined in the guide.
 
-> Due to what appears to be a bug in the Kubernetes Windows build system, one has to first build a Linux binary to generate `_output/bin/deepcopy-gen`. Building to Windows w/o doing this will generate an empty `deepcopy-gen`.
+    $ mkdir ~/kube
+    $ git clone https://github.com/Microsoft/SDN /tmp/k8s 
+    $ mv -R /tmp/k8s/Kubernetes/linux/ ~/kube/
```
JMesser81: We need to add a command to do this for windows scripts as well, correct? I think the section below on setting up windows nodes expects that kube binaries *and scripts* from SDN repo are in ~/kube-win/ dir.

PR #140, opened 2017-11-04
Get-NetView: run some functions in background threads

tmolenh:
## Overview
Use Runspaces to run the 3 slowest functions in background threads. For me this, this halved the execution time:
- 2m 20s → 1m 10s (-50%) on client
- 9m 48s → 4m 53s (-50.17%) on server

On systems with a lot of vSwitches, etc it's expected that the time improvement would be much less.

## Notes

- Closing the PowerShell window will just terminate any background threads.
- Ctrl-C works, but may take some time to stop all the background threads.


omarcardona:
Awesome.  This is great Tom!  I sent you some comments offline by diffing the RAW files since the GitHub diff is way too noisy for this change.  Please review and address comments.

PR #142, opened 2017-11-08
Get-NetView: fix missing quotes in some commands

tmolenh:
So they don't fail if the variable has spaces.


omarcardona:
How was it that the double-quotes for stripped?  All prior implementations were double-quotes specifically for the spacing reason?  If they got stripped by some editor setting or automation, then let's look into what caused it such that it doesn't occur again.

PR #143, opened 2017-11-09
Get-NetView: Call NicVendor per Nic

tmolenh:
1. Call NicVendor per Nic
2. ~~Pass Get-NetAdapter output to NicVendor for IHV convenience, and to avoid extra Get-NetAdapter calls.~~
3. Bug fixes

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -174,6 +174,7 @@ function Start-Thread() {
     [CmdletBinding()]
     Param(
         [parameter(Mandatory=$true)] [ScriptBlock] $ScriptBlock,
+        [parameter(Mandatory=$false)] [ValidateScript({Test-Path $_ -PathType Container})] [String] $StartPath = ".",
         [parameter(Mandatory=$false)] [Hashtable] $Params = @{}
     )
 
@@ -181,6 +182,7 @@ function Start-Thread() {
 
     $ps.Runspace = [RunspaceFactory]::CreateRunspace()
     $ps.Runspace.Open()
+    $null = $ps.AddScript("Set-Location ""$(Resolve-Path $StartPath)""")
     $null = $ps.AddScript($ExecFunctions) # import into thread context
     $null = $ps.AddScript($ScriptBlock).AddParameters($Params)
 
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -840,36 +840,35 @@ function ChelsioDetail {
         [parameter(Mandatory=$true)] [String] $OutDir
     )
 
-    # This only needs to be run once, but this function is called per Nic
+
     $dir = (Join-Path -Path $OutDir -ChildPath "ChelsioDetail")
-    if (-not (Test-Path -Path $dir)) {
-        New-Item -ItemType Directory -Path $dir | Out-Null
 
-        # Collect Chelsio related event logs and miscellaneous details
-        $file = "ChelsioDetail-Eventlog-BusDevice.txt"
-        $out = (Join-Path -Path $dir -ChildPath $file)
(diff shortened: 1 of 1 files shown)


Diagnostics/Get-NetView.PS1
```diff
@@ -880,12 +879,14 @@ function ChelsioDetail {
...
 
-    $ifName = (Get-NetAdapter -InterfaceIndex $IfIndex).Name
+    $ifName = $Nic.Name
+    $ifIndex = $Nic.ifIndex
     $ifHwInfo = Get-NetAdapterHardwareInfo -Name "$ifName"
     $dirBusName = "BusDev_" + $ifHwInfo.Bus + "_" + $ifHwInfo.Device + "_" + $ifHwInfo.Function
     $dirBus = (Join-Path -Path $dir -ChildPath $dirBusName)
 
-    if (-not (Test-Path $dirBus)) { # run once
+    if (-not $Global:ChelsioRunOnce) { # expensive
```
tmolenh: Takes some time to complete, and generates a somewhat large file.

Diagnostics/Get-NetView.PS1
```diff
@@ -651,9 +651,9 @@ function NetAdapterWorkerPrepare {
     Write-Host ""
     Write-Host "Processing: $title"
     Write-Host "----------------------------------------------"
-    NetIpNic         -NicName $name -OutDir $OutDir
+    NetIpNic         -NicName $name -OutDir $dir
     NetAdapterWorker -NicName $name -OutDir $dir
-    NicVendor        -OutDir  $dir
+    NicVendor        -Nic $nic      -OutDir $dir
```
omarcardona: Use the same format as prior.  Specifically -NicName instead of -Nic

Diagnostics/Get-NetView.PS1
```diff
@@ -836,7 +836,7 @@ function NativeNicDetail {
 function ChelsioDetail {
     [CmdletBinding()]
     Param(
-        [parameter(Mandatory=$true)] [String] $IfIndex,
+        [parameter(Mandatory=$true)] [CimInstance] $Nic, # Get-NetAdapter output
```
omarcardona: Use -NicName here for consistency with other CMDLETs input parameters.  Similar to prior comment.

Diagnostics/Get-NetView.PS1
```diff
@@ -936,11 +936,11 @@ function ChelsioDetail {
         $out = (Join-Path -Path $dirBus -ChildPath $file)
         $OutFlash = (Join-Path -Path $dirBus -ChildPath $flashFile)
         [String []] $cmds = "cxgbtool.exe $ifNameVbd hardware sgedbg",
-                            "cxgbtool.exe $ifNameVbd hardware flash $OutFlash"
+                            "cxgbtool.exe $ifNameVbd hardware flash ""$OutFlash"""
         foreach ($cmd in $cmds) {
             ExecCommand -Command $cmd -Output $out
         }
-    }
+    } # -not $Global:ChelsioRunOnce
```
omarcardona: This comment should be on 942,; the closing IF() basic block, rather than line 943 which is the function ending line.  As is, the comment will be misleading.

Diagnostics/Get-NetView.PS1
omarcardona: ChelsioRunOnce can't be a global.  There may be multiple physical NICs in the system in which case each must have the command executed.  It's "globalness" is with regard to the number of Ports on a given ASIC.   I'd say that if this takes a long time, then use a parallel execution mechanism as already in use in the main function Get-NetView().

tmolenh: It also outputs 16 MB of files (~2 MB) compressed. Running it once per "$dirBusName" should be good enough.

PR #183, opened 2018-02-28
Removes HNS module cmdlets that conflict with existing in-box cmdlets.

gkudra-msft:
Also updates the manifest files to use the v1.9.1 images for the master.


gkudra-msft:
Done in #184.

PR #187, opened 2018-03-05
Fix typo in start-kubelet.ps1

davidjahn:
The function "Get-HnsNetworks" does not exist... this should be "Get-HnsNetwork".


JMesser81:
Thank. LGTM

PR #199, opened 2018-03-16
! Missing download in the flannel start script

artiomchi:
Seems like the l2bridge flannel start script forgets to download the `pause.ps1` script. It's used in the dockerfile, so without it the container build call will fail


artiomchi:
@JMesser81 I'm guessing this would be for you, as I saw you were the last one to touch these?

JMesser81:
Sorry for the LONG delay on this. Can you please rebase and re-submit?

JMesser81:
@artiomchi ^^

artiomchi:
No worries! Also, seems like things have changed since I opened this PR. Specifically, the file that was missing has since been deprecated and removed from the repo, which makes this PR outdated.

Oh well, at least I can test it again, and see if I can make the new version work!

PR #205, opened 2018-04-11
Kubelet pause image - use cmd/bat instead of powershell for lower memory usage

jbiel:
Use a batch script for the kubelet pause image instead of powershell for much lower memory usage per pod


jbiel:
This seems to cause high CPU usage for some reason. Re-working and will re-open in the future.

PR #214, opened 2018-05-21
Adding hyperv container support for windows node.

rjaini:
Starting Kubelet binary accordingly based on isolationtype


madhanrm:
ping @JMesser81

PR #233, opened 2018-07-06
Fix scripts

madhanrm:
Update Scripts to check for file existence before downloading
Update Docs
Add VFP Helper to help dump vfpRules without Hyper-v Installed


madhanrm:
ping @JMesser81

PR #237, opened 2018-07-24
PS filtering fails with Get-HnsNetwork -Verbose argument

atomaras:
With -Verbose the result is not empty:
image

Without it it works:
image


daschott:
Thx, @atomaras I included this in: https://github.com/Microsoft/SDN/pull/282/files#diff-3f3a9a9ea4ef262dc51a675013054819R41

daschott:
Sorry for the long overdue response

PR #242, opened 2018-08-02
Fix running in remote powershell sessions

PatrickLang:
`Start-BitsTransfer` won't work through PowerShell remoting. `Invoke-WebRequest -UseBasicParsing` will. That's needed to get the first helper.psm1, then the DownloadFile calls will succeed


PatrickLang:
Wow, got a PR with one of my [favorite numbers](https://en.wikipedia.org/wiki/Front_242)

PR #249, opened 2018-08-31
Configuring kubelet logs redirect to same folder.

rjaini:
Configuring kubelet logs redirect to same folder. 
This will help in debugging issues in case of any failure.


madhanrm:
LGTM

PR #260, opened 2018-10-06
Updating scripts for Flannel Overlay

ksubrmnn:
Scripts have been updated to include Source VIP creation which is necessary to enable Service VIP connectivity. 

Environment variables are also set for Kube Proxy to properly create policies for service VIP. See [#69439](https://github.com/kubernetes/kubernetes/pull/69439).

This PR is being created so that it can be validated by others before merging. After that validation is complete, this PR will be updated to include up to date documentation.


ksubrmnn:
@benmoss

madhanrm:
@ksubrmnn Can you update the ReadMe to specify that this works only for RS5 and above?

PR #272, opened 2018-11-08
DNS zone permissions and MUX Routes

BronsonMagnan:
Added Update-SDNExpressDNSZonePermissions which handles network controller computer account permissions on the DNS Zone. Also Added Update-SDNExpressMuxRoutes, which
programs BGP peer routes as /32 routes and Other PA Sunet routes into the Muxes.


BronsonMagnan:
I think because of the DNS function, there is now a dependency on AD RSAT tools and PowerShell module on the machine that SDNExpress is executing on.

PR #281, opened 2018-11-20
Cleanup scripts

madhanrm:
Fix download script


daschott:
FYI just realized this might conflict with https://github.com/Microsoft/SDN/pull/282 which a few people are waiting on. Can you rebase it on top of that or is this urgent?

PR #311, opened 2019-02-04
Remove env variables from helper

ksubrmnn:
@madhanrm


daschott:
LGTM, I used this for validation.

PR #312, opened 2019-02-05
Unifying scripts

daschott:
* added install.ps1 for unifying Download-File calls and other setup procedure
* added parameters for configuring CIDRs to overlay start.ps1
* Added fix-up into helper.psm1 for non-default Mgmt NIC name

Open question: We could consolidate little bit further and add parameter $NetworkMode for start.ps1 and remove duplicate start.ps1, start-kubelet.ps1, start-kubeproxy.ps1. However, this would add more branches and conditionals. Is this desired/OK?

Outstanding work:
  * Add params to download & extract K8s binaries
  *  Add method to populate net-conf.json with customer CIDR


daschott:
@madhanrm @ksubrmnn Any preference for the open question above?

Kubernetes/windows/install.ps1
```diff
@@ -0,0 +1,84 @@
...
+{
+    Write-Host "Downloading Windows Kubernetes scripts"
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/hns.psm1 -Destination $BaseDir\hns.psm1
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/InstallImages.ps1 -Destination $BaseDir\InstallImages.ps1
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/Dockerfile -Destination $BaseDir\Dockerfile
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/stop.ps1 -Destination $BaseDir\stop.ps1
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/$NetworkMode/start-kubelet.ps1 -Destination $BaseDir\start-Kubelet.ps1
+    DownloadFile -Url  https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/$NetworkMode/start-kubeproxy.ps1 -Destination $BaseDir\start-Kubeproxy.ps1
+}
+
```
madhanrm: As part of a follow up PR, Can we also add a method to download the kubernetes binary? It would reduce one more manual step

madhanrm: Also the current scripts are updated which uses features from latest kubernetes.
Am not sure if we have to maintain scripts for older kubenetes version. eg. kube proxy changes

daschott: True, I removed the WinDSR feature gate. For overlay, this is a new feature we are advertising with v1.14 so it should be OK to keep it.

Kubernetes/flannel/overlay/README.md
```diff
@@ -4,8 +4,9 @@
 * Build Kube Proxy from [PR 70896](https://github.com/kubernetes/kubernetes/pull/70896) and copy to c:\k
 * Copy Kubeconfig from Linux master to c:\k
 * Download the following files to c:\k
-    [start.ps1](https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/overlay/start.ps1) 
-    [helper.psm1](https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/helper.psm1) 
+  * [start.ps1](https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/overlay/start.ps1)
```
bclau: you probably mean ``https://github.com/Microsoft/SDN/raw/master/Kubernetes/flannel/start.ps1``, because the file you mention will be deleted in this PR.

Kubernetes/windows/start-kubeproxy.ps1
```diff
@@ -1,9 +1,10 @@
 Param(
```
bclau: So, there are going to be 2 start-kubeproxy.ps1 files. When should we use either of them?

Plus, there are also ``start-kubelet.ps1``, ``start.ps1``, ``stop.ps1`` in the windows folder. Are we not going to remove them?

Kubernetes/flannel/start-kubeproxy.ps1
```diff
@@ -0,0 +1,22 @@
...
+    $NetworkName = "cbr0"
+)
+
+$networkName = $NetworkName.ToLower()
+ipmo c:\k\hns.psm1
+Get-HnsPolicyList | Remove-HnsPolicyList
+
+if ($NetworkMode -eq "l2bridge")
+{
+    c:\k\kube-proxy.exe --v=4 --proxy-mode=kernelspace --hostname-override=$(hostname) --kubeconfig=c:\k\config --network=$networkName --enable-dsr=false --log-dir=$LogDir --logtostderr=false
```
bclau: isn't it ``--network-name``? kube-proxy doesn't recognize the ``--network`` argument, and you're also passing ``--network-name`` below for overlay.

Kubernetes/flannel/start.ps1
```diff
@@ -0,0 +1,75 @@
+﻿Param(
+    [ValidateSet("l2bridge", "overlay",IgnoreCase = $true)] [parameter(Mandatory = $true)] $NetworkMode,
+    [parameter(Mandatory = $false)] $ClusterCIDR="10.244.0.0/16",
+    [parameter(Mandatory = $false)] $KubeDnsServiceIP="10.96.0.10",
+    [parameter(Mandatory = $false)] $ServiceCIDR="10.96.0.0/12",
+    [parameter(Mandatory = $false)] $InterfaceName="Ethernet",
+    [parameter(Mandatory = $false)] $LogDir = "C:\k",
+    [parameter(Mandatory = $true)] $ManagementIP
```
bclau: So, we're not going to have $IsolationType anymore?

Kubernetes/flannel/overlay/README.md
daschott: Yes, these need updating indeed. I was going to update the readme.md's after we finalized the decision where we want to keep the files (or revert the previous commit alternatively)

ksubrmnn:
@daschott  Regarding consolidating scripts and taking a network type parameter, I think that I'm a bit partial to having separate scripts since it makes it easier to debug since it's straightforward what you're passing in. Maybe start kubelet can be just one file? Not sure what differences are there

ElanHasson:
Is there a plan to start using the windows service functionality already in kubelet and kubeproxy?

daschott:
@ElanHasson There is some docs on this here: https://kubernetes.io/docs/getting-started-guides/windows/#kubelet-and-kube-proxy-can-now-run-as-windows-services

Otherwise In the next docs.microsoft.com update since [nssm.exe](https://nssm.cc/download) can be simpler to start with and it is used on AKS-Engine we will include a section (and possibly a sample configuration script) to register flanneld.exe, kube-proxy.exe, and kubelet.exe as services. I know some users already have done that e.g. to register kubelet service:
```
nssm install kubelet C:\k\kubelet.exe

nssm set kubelet AppParameters --hostname-override=<hostname> --v=6 --pod-infra-container-image=kubeletwin/pause --resolv-conf="" --allow-privileged=true --enable-debugging-handlers --cluster-dns=<DNS-service-IP> --cluster-domain=cluster.local --kubeconfig=c:\k\config --hairpin-mode=promiscuous-bridge --image-pull-progress-deadline=20m --cgroups-per-qos=false  --log-dir=c:\k\logs --logtostderr=false --enforce-node-allocatable="" --network-plugin=cni --cni-bin-dir=c:\k\cni --cni-conf-dir=c:\k\cni\config

nssm set kubelet AppDirectory C:\k

nssm start kubelet
```

daschott:
@ksubrmnn Thanks! Only trade-off for that is that we need to maintain both scripts against any changes and keep them in sync. But definitely happy to revert the commit if it's the lesser of two evils! Start-Kubelet.ps1 I think there are differences because [populating CNI config](https://github.com/Microsoft/SDN/pull/312/files#diff-3f3a9a9ea4ef262dc51a675013054819R241) is different. If we remove that call and ask users to populate cni.conf, then we could have a single start-kubelet.ps1 I believe.

@madhanrm @ElanHasson any opinions on consolidating the scripts? Perhaps I'll simply defer to end-users during next SIG-Windows call as well if they have any preference and ultimately close on this.

Kubernetes/flannel/start.ps1
ElanHasson: Yes-- I'm curious about this also. What about support for the `-feature-gates=HyperVContainer=true` flag?

ElanHasson:
Personally, I don't want to populate a CNI config unless I have to.

I think a level of consolidation is necessary and possible without sacrificing the implementation details that each network mode brings. I look at this and see this as an interface with multiple implementations problem.

We could switch to a plug-in based model, depending on the value of `$NetworkMode` for different support functions.

Load `"$(NetworkMode).psm1"` which implements the following _interface_:

|Function|Description|
|---|---|
|GetCniConfig | Takes in command line parameters to generate CNI Config. Returns an object to write as Json to `cni.config`
|GetFilesToDownload| Returns an array of objects in shape of `{sourceUrl, targetRelativePath}`. 
|ResolveKubeProxyArgs| Takes in an array of key/value pairs representing the flag and the value. Returns a list of key-value pairs with removals/additions based on the implementation.
|ResolveKubeletArgs| Takes in an array of key/value pairs representing the flag and the value. Returns a list of key-value pairs with removals/additions based on the implementation.
|GetNetworkName|Returns a string used to populate `$NetworkName`
|CreateNetwork|Creates any HNS work and starts the network (i.e. Starts Flanneld)


I'm sure I missed a few functions on the interface, but I think it conveys the point.

The orchestration of the network plugins will allow for them to evolve independently while the flow remains the same, where common code is still shared and they still remain testable and debuggable on their own.

Above all, no more `if ($NetworkMode -eq "blah")` scattered throughout the code.

ElanHasson:
@daschott @ksubrmnn any feedback on my comment? 
I can do a PR if you like...

Kubernetes/flannel/start.ps1
daschott: Are you using this and does it work for you? This was originally added purely for consistency with the other start.ps1 scripts in this repo, but there was some issues with endpoint attach since it was in alpha. These scripts are intended for customers to be a reference to get started, then they can customize it to their liking by adding flags for custom alpha features like DSR or hyper-V pods which should be documented separately.

The only exception for this right now is the overlay feature, which is a network mode we want customers to be able to select easily vs. l2bridge.

Kubernetes/flannel/start-kubeproxy.ps1
daschott: Thanks, I validated l2bridge now & fixed the typo.

Kubernetes/windows/start-kubeproxy.ps1
daschott: The scripts in the windows folder are for wincni directly without flannel delegation. Right now, they are not in the mainline docs so they shouldn't be used unless there is a specific requirement.

daschott:
@ElanHasson Sure, that sounds reasonable. Feel free to raise a PR and ping us, thanks!

madhanrm:
@daschott  is this ready for merge?

daschott:
@madhanrm Almost, I am pending validation results for l2bridge on RS4 and older K8s versions. I should also add a clarification that VSID and port needs to be specified for Linux Flannel overlay .yml spec. I will ping you when ready to merge.

ksubrmnn:
@ElanHasson Sorry I didn't get back to you sooner! Ping us when your PR is ready. Thanks for contributing!

ElanHasson:
No worries @ksubrmnn. I'm out of the office this week, I should be able to work on it this in the next week.

Which script folders should be consolidated? All of them?

PR #314, opened 2019-02-13
Validate Kubernetes Scripts

ksubrmnn:
@madhanrm

### DIFF
--- a/Kubernetes/windows/test/ValidateKubernetes.Pester.tests.ps1
+++ b/Kubernetes/windows/test/ValidateKubernetes.Pester.tests.ps1
@@ -1,0 +1,122 @@
+<#
+.DESCRIPTION
+    This script validates basic connectivity for Windows nodes in a Kubernetes cluster. 
+    Assumes the cluster has no other services/pods deployed besides kube-system
+#>
+
+import-module "$PSScriptRoot\ValidateKubernetesHelper.psm1"
+
+Describe 'Kubernetes Prerequisites' {
+    
+    Context 'Checking Docker images' {
+        It "should have nanoserver image" {
+            docker images | findstr nanoserver | Should Not BeNullOrEmpty
+        }
+        It "should have windowservercore image" {
+            docker images | findstr windowsservercore | Should Not BeNullOrEmpty
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Kubernetes/windows/debug/ValidateKubernetesHelper.psm1
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return $containers;
+}
+
+function TestConnectivity()
+{
+    param(
+        [string] $containerName,
+        [string] $remoteHost =  "172.217.3.206", #google.com
```
madhanrm: Doens;t name work here? That would validate DNS as well

Kubernetes/windows/debug/ValidateKubernetesHelper.psm1
```diff
@@ -0,0 +1,127 @@
...
+    throw "TCP connection to ${remoteHost}:${port} failed from $containerName."
+}
+
+function GetContainerIPv4Address()
+{
+    param(
+        [string] $containerName
+    )
+
+    $matches = (kubectl exec $containerName ipconfig | Out-String  | Select-String -Pattern '(?sm)(IPv4 Address).*?: (.*?)\r\n' -AllMatches).Matches
```
madhanrm: can you also add a test to validate if kubectl get pod <podname> is reporting the same podip as the container ip?

madhanrm:
This is cool. I'll merge this and we can add more tests.

madhanrm:
This is cool. I'll merge this and we can add more tests.

can you move the files under windows/test  instead of windows/debug?

daschott:
Awesome, thanks for doing this. I will try it out and see if it makes sense to add to next docs update. I'm sure it would be useful for users doing validation as well.

Kubernetes/windows/debug/ValidateKubernetesHelper.psm1
ksubrmnn: This test compares ipconfig with kubectl reported ip

ksubrmnn:
@madhanrm moved the files. I'll include a DNS check when I use the updated CNI and follow up in a separate PR

PR #316, opened 2019-02-19
Replacing Start-BitsTransfer Call

celestialorb:
Start-BitsTransfer fails when used in userdata in an AWS EC2 instance since no user is logged in (not sure why this is the case). Replaced the call with a call to `wget`, but I'm not a Powershell expert, perhaps an `Invoke-WebRequest` might be better suited here.


daschott:
Thx for raising this -- can you rebase this so that it can be merged?

PR #317, opened 2019-02-19
Better Flannel Startup

celestialorb:
Adding a retry loop around the flannel startup as it seems to often times get stuck at "waiting for network to be created" which can be solved by simply killing it and restarting the job.

Added a hardcoded 60 second timeout for flannel, could be supplied as a parameter.


daschott:
Can this be rebased? This workaround is a good temporary measure in the scripts while the underlying issue is being investigated.

FYI @madhanrm @ksubrmnn can one of you merge this once rebased?

ksubrmnn:
@JocelynBerrendonner another instance of needing flanneld restart

JocelynBerrendonner:
@dineshgovindasamy FYI

PR #318, opened 2019-02-19
Node Role Label

celestialorb:
Adding an optional node role label parameter. Fixed the issue of the node showing up in Kubernetes with a role of `<none>`.


daschott:
Can you rebase this please? Then we can merge it.

PR #327, opened 2019-03-07
Get-NetView: PowerShell Core compat

tmolenh:
1. Replace cmdlets unsupported by PowerShell Core, while maintaining PowerShell 5 compatibility:
   - WMI cmdlets with CIM cmdlets.
   - Get-EventLog with Get-WinEvent.
   - Get-Counter has no replacement at this time, so I rewrote the counter collection to use typeperf instead.
2. Replaced wmic, since it is deprecated.


omarcardona:
Please adjust the version/date for this update.

PR #335, opened 2019-04-02
Fixes for Kubernetes 1.14 windows node setup

PatrickLang:
I ran into two problems following the guide today:

1. `start.ps1` was throwing an error that `Argument '2' …` was not expected. This was because my Ethernet adapter was `Ethernet 2` which includes a space. The network adapter name needs to be quoted to handle spaces.
2. I was getting an OS mismatch starting pods. Moving to the published multi-version pause image fixes this and speeds up setup.

### DIFF
--- a/Kubernetes/flannel/start-kubelet.ps1
+++ b/Kubernetes/flannel/start-kubelet.ps1
@@ -28,7 +28,7 @@ if ($RegisterOnly.IsPresent)
 $kubeletArgs = @(
     "--hostname-override=$(hostname)"
     '--v=6'
-    '--pod-infra-container-image=kubeletwin/pause'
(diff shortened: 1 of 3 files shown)


PatrickLang:
@ksubrmnn or @madhanrm can you review?

ksubrmnn:
Oh cool. Didn't know we had multi-version images published. LGTM

daschott:
LGTM, I validated these changes. Thanks for the change for the mult-arch image.  @madhanrm can you merge this?
 
@PatrickLang also FYI Madhan did work on improved scripts that makes configuration simpler: https://github.com/Microsoft/SDN/tree/master/Kubernetes/windows/kubeadm we should document them too. And then later update everything with kubeadm

PR #360, opened 2019-05-08
Fixes to improve the reliability of setup scripts.

rjaini:
Fixes to improve the reliability of setup scripts.


rjaini:
@madhanrm

PR #366, opened 2019-05-16
Replace non-existing pause-image name

asbjorne:
The script refers to a image not longer hosted in the registry. Replaced imagename with the same used when running start-kubelet from start.ps1 (https://raw.githubusercontent.com/Microsoft/SDN/master/Kubernetes/flannel/start-kubelet.ps1)


jonaskello:
This solution works! For now you can manually patch [this line](https://github.com/microsoft/SDN/blob/master/Kubernetes/flannel/register-svc.ps1#L25) inregister-svc.ps1  to use `--pod-infra-container-image=mcr.microsoft.com/k8s/core/pause:1.0.0`.

dhirschfeld:
ping!

PR #370, opened 2019-06-05
Get-NetView: Improve error resilience

tmolenh:
1. Make sure Sanity and Completion are always executed, so a zip is created.
2. If the script crashes, the error message is collected and zipped.
3. Set parameter mandatory = $false for some parameters. This allows them to be null or empty, which should prevent or delay some crashes.
4. Remove `-Trusted` commands since there is no longer a performance benefits, and it can hide errors from the output.

#### Other Changes
1. New function ExecControlError for consistent error logging.
2. Rename NetNat to NetNatDetail.
3. Combine HwErrorReport and LogsReport into one function.
4. Slightly reorder some modules based on importance.

### DIFF
--- a/Diagnostics/Get-NetView.PS1
+++ b/Diagnostics/Get-NetView.PS1
@@ -2093,20 +2093,7 @@ function Counters {
     typeperf -f BIN -o $out -sc 10 -si 5 $counterPaths > $null
 } # Counters()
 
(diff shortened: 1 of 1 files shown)


tmolenh:
Closing because I added this PR here: https://github.com/microsoft/Get-NetView/pull/7

PR #378, opened 2019-07-04
Fix vmmexpress timing

adamfd:
Modified:
NC template to add a step after the reboot to ensure the network stack has come up before it tries to log in with domain account - frequent issues that the script would execute before the network was up.
SLB template to add a step after the reboot to wait - numerous failures where a box would reboot and after it came up the other hosts would fail.
VMMExpress.ps1 - added a loop to wait for the network controller service to return as we had a regular failure where the script moved too fast the the variable was null


adamfd:
Added a further update to the GetCertChain function in the SLB processcertificates.ps1 file.
This was causing random powershell crashes on both 2016 and 2019.
Replaced fir a dir cert instead to achieve the same thing.

Issue #379, opened 2019-07-09
Remove flag "--allow-privileged" form start-kubelet.ps1 in Kubernetes 1.15


mrwonkerz:
The start-kubelet.ps1 runs in the an error when using with kubernetes version 1.15
In Kubernetes 1.15 the flag "--allow-privileged" for kubelet is obsolete.

```
.\start-kubelet.ps1 -NetworkMode Overlay -KubeDnsServiceIP 10.96.0.10 -LogDir C:\k\Logs
Flag --enable-debugging-handlers has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Usage:
  kubelet [flags]
...
F0709 11:25:37.135712    3088 server.go:156] unknown flag: --allow-privileged
```

The kubelet-Service can not be started because of this.
Removing the flag "--allow-privileged" from the start-kubelet.ps1 script fixes the issue.

```
$kubeletArgs = @(
    "--hostname-override=$(hostname)"
    "--v=6"
    "--pod-infra-container-image=mcr.microsoft.com/k8s/core/pause:1.0.0"
    "--enable-debugging-handlers"
#   "--allow-privileged=true"
    "--cluster-dns=$KubeDnsServiceIp"
    "--cluster-domain=cluster.local"
    "--kubeconfig=c:\k\config"
    "--hairpin-mode=promiscuous-bridge"
    "--image-pull-progress-deadline=20m"
    "--cgroups-per-qos=false"
    "--log-dir=$LogDir"
    "--logtostderr=false"
    "--enforce-node-allocatable="""
    "--network-plugin=cni"
    "--cni-bin-dir="c:\k\cni""
    "--cni-conf-dir="c:\k\cni\config""
    "--node-ip=$(Get-MgmtIpAddress)"
)
if ($KubeletFeatureGates -ne "")
{
    $kubeletArgs += "--feature-gat
```

wk8:
Fixed in https://github.com/microsoft/SDN/pull/383

mrwonkerz:
@wk8 I found out that the same problem exists int the register-svc.ps1 script.

Thanks for the fix

wk8:
@mrwonkerz : thanks, good catch. Should be fixed by the new commit I've added to my PR (https://github.com/microsoft/SDN/pull/383/commits/0965379335cd58aaef247d6d8b6699d04aec6e79).

PR #383, opened 2019-07-10
Only using the kubelet flag `allow-privileged` for versions < 1.15

wk8:
Since that flag got deprecated in version 1.15, and the kubelet won't
start if it's present.

Fixes https://github.com/microsoft/SDN/issues/379

### DIFF
--- a/Kubernetes/flannel/register-svc.ps1
+++ b/Kubernetes/flannel/register-svc.ps1
@@ -9,6 +9,19 @@ Param(
     [parameter(Mandatory = $false)] $FlanneldSvc="flanneld"
 )
 
+$GithubSDNRepository = 'Microsoft/SDN'
+if ((Test-Path env:GITHUB_SDN_REPOSITORY) -and ($env:GITHUB_SDN_REPOSITORY -ne ''))
+{
+    $GithubSDNRepository = $env:GITHUB_SDN_REPOSITORY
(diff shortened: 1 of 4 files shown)


dhirschfeld:
It would be great if this could be merged since 1.15 has been released a month ago.

I've always found it better to merge quickly and iterate. If there are new bugs they'll be fixed quickly due to the increased pace of development and they can't be worse than the current situation of being completely broken for k8s 1.15.

dhirschfeld:
ping!

szb640:
Any update on the pull request? Following the steps in the [Kubernetes' official guide](https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-nodes/#components-that-run-on-windows) are not working, but should:
 > Get the latest binaries from https://github.com/kubernetes/kubernetes/releases, starting with v1.14 or later

wk8:
Maybe @madhanrm could take a look? :)

madhanrm:
@daschott  can you take a look?

daschott:
@madhanrm  Sure, I heard this is set in config file like the comment indicates. I will validate now.

daschott:
LGTM, validated on 1.13 and 1.15. @madhanrm can you merge?

wk8:
Thank you! :)

PR #385, opened 2019-07-17
Remove invalid setting for 1.15

sixeyed:
`allow-privileged` causes the kubelet to fail in 1.15. Remove this and the script works correctly - there are still various other deprecated settings which throw warnings, but this is the only one which stops the kubelet running.


dhirschfeld:
looks like a dup of #383

sixeyed:
True, and that one is a better implementation :)

PR #388, opened 2019-08-05
Support kubelet 1.15 on Windows

szb640:
Starting with version 1.15 the --allow-privileged switch has been [deprecated](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#node) in kubelet. Because of this `kubelet.exe` is not started when running the script referenced in the [Kubernetes Windows Guide](https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-nodes/#join-the-windows-node-to-the-flannel-cluster).

### DIFF
--- a/Kubernetes/windows/start-kubelet.ps1
+++ b/Kubernetes/windows/start-kubelet.ps1
@@ -116,7 +116,7 @@ if ($IsolationType -ieq "process")
 {
     c:\k\kubelet.exe --hostname-override=$(hostname) --v=6 `
         --pod-infra-container-image=kubeletwin/pause --resolv-conf="" `
-        --allow-privileged=true --enable-debugging-handlers `
+        --enable-debugging-handlers `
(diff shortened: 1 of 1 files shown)


dhirschfeld:
I assume this was closed because it is superseded by #383

szb640:
> I assume this was closed because it is superseded by #383

Yes, I noticed it too late, however the added extra logic in that PR makes me wonder whether a smaller change like this would be merged faster. According to the documentation the default value of `--allow-privileged` was `true` by default, so not sure if the flag is necessary at all before 1.15.

dhirschfeld:
I'm not sure how to get that PR merged faster as no-one has engaged in the discussion over there :(

I've linked this PR but maybe if you (@szb640) want to comment there it may lend weight/credence to the fact that it is an actual problem for anyone wanting to use k8s on Windows.

PR #392, opened 2019-08-12
Pass Release parameter in to InstallKubernetes function

superlime:
Currently kubecluster.sh doesn't actually respect the release specified when picking K8S packages to install from APT.  This PR filters the makes it select from only packages matching the supplied release string.


madhanrm:
LGTM

PR #398, opened 2019-09-07
Adding port reservations and Windows version to collectlogs

JocelynBerrendonner:
Adding port reservation and windows version to collectlogs.ps1

Tested: Manually tested the change on Server SKU and Client SKU

### DIFF
--- a/Kubernetes/windows/debug/collectlogs.ps1
+++ b/Kubernetes/windows/debug/collectlogs.ps1
@@ -65,5 +65,21 @@ if ($res)
     docker ps -a > docker.txt
 }
 
+netsh int ipv4 sh excludedportrange TCP > reservedports.txt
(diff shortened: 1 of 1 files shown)


Kubernetes/windows/debug/collectlogs.ps1
```diff
@@ -65,5 +65,21 @@ if ($res)
     docker ps -a > docker.txt
 }
 
+netsh int ipv4 sh excludedportrange TCP > reservedports.txt
+netsh int ipv4 sh excludedportrange UDP >> reservedports.txt
+
```
madhanrm: Is ICMP also valid?

JocelynBerrendonner: No, it isn't; I have just checked. I don't think we can reserve any ports for ICMP.

PR #426, opened 2020-04-03
Bug fixes for join mode

k-modi:
I was experiencing issues when trying to use the ```Kubernetes\linux\kubeadm\kubecluster.sh``` script to join my ubuntu worker nodes to the cluster. I'm using the same script to initialize the master control-plane and it works great. Hence, I decided to investigate why join was not working smoothly. In the process, I discovered some issues as explained below and would like to request a review and if all looks good, then merge the changes.

There were basically 2 main issues:

1. The script is executing ```kubeadm join``` command to join the node. However, the kube config file was not being copied over to ```$HOME/.kube/config``` directory. I noticed that there is the ```InstallKubeConfig``` function defined and it is copying the ```admin.conf``` from ```/etc/kubernetes```. Hence, I decided to update the function to also support ```join``` mode and alternatively copy the ```kubelet.conf``` from the same location as ```$HOME/.kube/config```. I also, added a call to this function after ```kubeadm join``` call from the Main script section. Without this, the script was subsequently failing when executing ```kubectl``` commands as there was no way for it to know the master configuration.
2. I also noticed the although, Installing and configuring network plugins is only required when initializing the master node, it was being done even when joining the cluster. This seems redundant and even if we try to do it, it results in script errors because the master plane throws back permission denied errors when trying to get object from the ```kube-system``` ns. Hence, I refactored this to be only called when initializing the master node.
3. Lastly, I noticed that beyond all the things ```InstallNetworkPlugins``` function was doing it was also updating the setting ```sudo sysctl net.bridge.bridge-nf-call-iptables``` to ```1```. I wasn't sure if it is really required on the worker node as well, so I'm adding the change with this merge request. If not needed, I can remove it.


k-modi:
Hello,

It's been a few days since this PR was created. can you please give some feedback or accept the PR? It will help us out either way.

k-modi:
Hello,

Just a friendly reminder, please review this PR and merge or provide a feedback so that further actions can be determined. 

Your cooperation is much appreciated! Thanks!

PR #458, opened 2020-12-22
Add NetMgmt, VMSIF and VmSwitch (control only) providers

Keith-Mange:
The VmSwitch provider uses keywords to limit it to control path only.

I'm adding Microsoft.Windows.Hyper.V.VmsIf {6C28C7E5-331B-4437-9C69-5352A2F7F296} and the control path events for VmSwitch ETW and WPP {67DC0D66-3695-47c0-9642-33F76F7BD7AD} and {1F387CBC-6818-4530-9DB6-5F1058CD7E86}.


mkostersitz (approved):
LGTM

sbangari (approved):
I will stop using this script in the favor of  starthnstrace.ps1 which looks much cleaner :) . Thanks for adding that.

PR #462, opened 2021-01-06
Copy windows event logs as part of CollectLogs output

sbangari:


Kubernetes/windows/debug/collectlogs.ps1
```diff
@@ -166,5 +166,11 @@ if ($hotFix -ne $null)
     "<No hotfix>" >> winver.txt
 }
 
+# Copy the Windows event logs
+Copy-Item "$env:SystemDrive\Windows\System32\Winevt\Logs\Application.evtx"
+Copy-Item "$env:SystemDrive\Windows\System32\Winevt\Logs\System.evtx"
+Copy-Item "$env:SystemDrive\Windows\System32\Winevt\Logs\\Microsoft-Windows-Hyper-V*.evtx"
+Copy-Item "$env:SystemDrive\Windows\System32\Winevt\Logs\\Microsoft-Windows-Host-Network-Service*.evtx"
```
erfrimod, line 173: Is it worth adding `-ErrorAction SilentlyContinue` to the end to prevent errors if these logs are not found? They all should exist, so it's probably not needed.

erfrimod (commented):
Looks good to me

Keith-Mange (commented):
LGTM.  Thanks for adding these -- they will be useful.

Kubernetes/windows/debug/collectlogs.ps1
sbangari: Copy-Item doesn't seem to complain or throw an error when it doesn't find the specified file. So I will it like this for now

PR #474, opened 2021-02-19
adding k8s-windows-debug-PowerShell_Updates

JamesKehr:
Rebuild of starthnstrace.ps1 and startpacketcapture.ps1.

### DIFF
--- a/Kubernetes/windows/debug/starthnstrace.ps1
+++ b/Kubernetes/windows/debug/starthnstrace.ps1
@@ -1,43 +1,311 @@
-param(
-    [bool] $Prompt = $true,
-    [string] $EtlFile = "c:\server.etl"
+[CmdletBinding()]
+param
+(
+    # Path with filename where the ETL file will be saved. Format: <path>\<filename>.etl
+    [string]
+    $EtlFile = "C:\server.etl",
+
+    # How many bytes of the packet to collect. Default is 256 bytes to collect encapsulated headers.
+    [int]
+    $snapLen = 256,
+
+    # Does not prompt/pause execution and wait on user input.
+    [switch]
+    $NoPrompt,
+
+    # Does not collect network packets.
+    [switch]
+    $GetPackets,
+
+    # Collects logs after user presses q to stop tracing. Ignored when -NoPrompt set.
+    [switch]
+    $CollectLogs
 )
 
-Stop-NetEventSession HnsCapture -ErrorAction Ignore
-Remove-NetEventSession HnsCapture -ErrorAction Ignore
+### CLASSES AND FUNCTIONS ###
+#region
 
-New-NetEventSession HnsCapture -CaptureMode SaveToFile -LocalFilePath $EtlFile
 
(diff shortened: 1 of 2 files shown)


JamesKehr:
- Modified the params of each script with additional options.

starthnstrace.ps1
-- 

    # How many bytes of the packet to collect. Default is 256 bytes to collect encapsulated headers.
    [int]
    $snapLen = 256,

    # Does not prompt/pause execution and wait on user input.
    [switch]
    $NoPrompt,

    # Collect network packets.  <<< need to edit this comment post-PR
    [switch]
    $GetPackets,

    # Collects logs after user presses q to stop tracing. Ignored when -NoPrompt set.
    [switch]
    $CollectLogs


startpacketcapture.ps1
--

    # How many bytes of the packet to collect. Default is 256 bytes to collect encapsulated headers.
    [int]
    $snapLen = 256,

    # Does not prompt/pause execution and wait on user input.
    [switch]
    $NoPrompt,

    # Does not collect network packets.
    [switch]
    $NoPackets,

    # Collects logs after user presses q to stop tracing. Ignored when -NoPrompt set.
    [switch]
    $CollectLogs


- Added [CmdletBinding()] to allow Verbose and Debug outputs.
- Added Write-Verbose and Write-Debug output for troubleshooting purposes.
- startpacketcapture collects packets by default. Can be disabled with -NoPackets
- starthnstrace does not collect packets by defaul. Can be enabled with -GetPackets
- Created [Provider] classes to manage ETW providers.
- Created Get-WebFile to download CollectLogs.ps1, if needed.
- [Provider[]]$providers is a centralized array of providers. Modifying this array to changes the captured data without hunting around for existing providers and messing with individual add-neteventprovider commands.
- Added try-catch logic to key NetEventSession calls. Prevents the script from continuing if there errors during execution.
- The Catch logic returns a terminating error. This allows a script calling start*.ps1 to terminate on error.
- Added logic to download and execute collectlogs.ps1. This is enabled by adding -CollectLogs to the script execution.
- Modified end of execution logic with location of data collected.

Keith-Mange (commented):
LGTM.  I appreciate your help on this.

Kubernetes/windows/debug/starthnstrace.ps1
```diff
@@ -1,43 +1,311 @@
...
+    [int]
+    $snapLen = 256,
+
+    # Does not prompt/pause execution and wait on user input.
+    [switch]
+    $NoPrompt,
+
+    # Does not collect network packets.
+    [switch]
+    $GetPackets,
```
sbangari, line 18: nit: can we change this to "NoPackets" as in the other script startpacketcapture.ps1

sbangari:
@Keith-Mange  I think we should change the name of starthnstrace.ps1 to something more meaningful because it captures more than just the HNS traces now.

sbangari (approved):

Keith-Mange:
> @Keith-Mange I think we should change the name of starthnstrace.ps1 to something more meaningful because it captures more than just the HNS traces now.

Sounds reasonable

PR #510, opened 2021-11-30
Add code for Monitoring Windows AKS Node for failures

sbangari:
Problem:
We have had ICMs in the recent past where Windows worker nodes on AKS were running into problems randomly during the course of their execution. To investigate these issues we need to collect logs/traces/events/dumps from the node immediately after the failure. This process has been cumbersome with challenges ranging from:
1. How do I detect the node that went into a bad a state?
2. How do I collect the logs? What commands/scripts to run, when and how? This needs to be communicated via an additional hop through an escalation engineer. Multiple cycles were wasted just in communicating the detailed steps across to the customer.
3. Monitoring the nodes for failure has been a manual process.

As a first step to solve this problem:

1. The developer implements a couple of functions translating his intent into a few lines of code.
#Implement these 4 methods:
# LogMessage - (Optional) Implements logic to log messages. Defaults to logging to a file.
# StartHandler - (Optional) Handler invoked after the monitoring starts (before the node is in repro state)
# TerminateHandler - (Optional) Handler invoked before the monitoring stops (after the node is in repro state)
# IsNodeFaulted - Returns a $true when the node is in repro state, $false otherwise

2. Then a scheduled task is created on the nodes using the "Monitoring Strategy" which was defined above.
example: 
Register-ScheduledJob -Name "MonitorWindowsNode" -FilePath C:\k\debug\MonitorWindowsNode.ps1 -RunNow -ArgumentList "C:\k\debug\LoadBalancerPolicyStrategy.psm1"

3. When the node goes into a faulted state as described by the strategy, logs are collected from the node. I also have a sample where these logs are automatically uploaded to Azure Blob storage.

### DIFF
--- a/Kubernetes/windows/debug/monitoring/MonitorWindowsNode.ps1
+++ b/Kubernetes/windows/debug/monitoring/MonitorWindowsNode.ps1
@@ -1,0 +1,118 @@
+[CmdletBinding()]
+param
+(
+    # Path to the module defining the strategy to use for monitoring the node
+    [string]
+    $StrategyModulePath = "C:\k\debug\StrategyModulePath.psm1"
+)  
+
+function Start-HNSTrace
+{
+    .\collectlogs.ps1
+    $sessionName = 'HnsCapture'
+    Write-Host "Starting HNS tracing"
+
+    $curDir = Get-Location
+    # Generate a random directory to capture all the logs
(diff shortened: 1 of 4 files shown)


daschott:
@sumicalbin Did you takeover this PR and merge it? Can this be closed?

sumicalbin:
@daschott No, I added a partial update. sbangari created comments for improvements that I cannot see for some reason.

PR #511, opened 2021-12-16
Verify elevated

Keith-Mange:
Warn users that attempt to run this without being elevated.

### DIFF
--- a/Kubernetes/windows/debug/collectlogs.ps1
+++ b/Kubernetes/windows/debug/collectlogs.ps1
@@ -1,12 +1,8 @@
+#Requires -RunAsAdministrator
+
 Param(
     [parameter(Mandatory = $false)] [string] $Network = "L2Bridge"
 )
-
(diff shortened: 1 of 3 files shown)


Kubernetes/windows/debug/collectlogs.ps1
```diff
@@ -2,6 +2,12 @@ Param(
     [parameter(Mandatory = $false)] [string] $Network = "L2Bridge"
 )
 
+if (-NOT ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator"))
```
daschott, line 5: I validated that this works. But would it perhaps be simpler if we replace the 4 lines by writing the following at the top of each of the scripts:
```
#Requires -RunAsAdministrator
```
Then we can rely on PowerShell's built-in error message.

Keith-Mange: Great suggestion, thank you!

daschott:
LGTM, I validated this works on my machine

PR #516, opened 2022-03-02
Add additional firewall information

Keith-Mange:


Kubernetes/windows/debug/collectlogs.ps1
```diff
@@ -111,7 +111,12 @@ Get-NetAdapter -IncludeHidden  | foreach {
     $_ | FL * | Out-File -Append -FilePath "adapters\$fileName" -Encoding ascii
     }
 
+New-Item -Path firewall -ItemType Directory
+cd firewall
 Get-NetFirewallRule -PolicyStore ActiveStore >> firewall.txt
+reg.exe query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\SharedAccess\Parameters\FirewallPolicy /s > fwpol.txt
+reg.exe query HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\WindowsFirewall /s > gppol.txt
```
daschott, line 118: This line produces an empty file `gppol.txt` for me. Can't find this path in the registry. It could be my environment though?

Keith-Mange: Interesting.  I did a bunch of output on mine.

houha2: From my experience, it will only have an output if rules have been applied through group policy. This is typically through some IT management software.

daschott: I tried it on Windows Client and Windows Server, but either way if it worked on other machines and depends on group policy LGTM

PR #531, opened 2022-08-18
add HCN namespaces to logs and print v2 representation by default

daschott:
* Add argument to collectlogs.ps1 to pick HNS schema version to represent in logs
* Collectlogs.ps1 imports HNSv2.psm1 
*  Make HNSv2.psm1 use v2 representation by default
* Add HCN Namespaces logging to collectlogs.ps1

### DIFF
--- a/Kubernetes/windows/debug/collectlogs.ps1
+++ b/Kubernetes/windows/debug/collectlogs.ps1
@@ -42,8 +42,13 @@ cd $outDir
 # V1
 if($HnsSchemaVersion -eq 1)
 {
+    # V1 Networks
     Get-HnsNetwork -Version 1 | Select Name, Type, Id, @{Name="AddressPrefix"; Expression={$_.Subnets.AddressPrefix}} > network.txt
+
+    # V1 Endpoints
     Get-HnsEndpoint -Version 1 | Select IpAddress, MacAddress, IsRemoteEndpoint, State > endpoint.txt
+
+    # V1 Namespaces
     Get-HnsNamespace -Version 1 `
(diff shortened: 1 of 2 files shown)


sbangari (approved):

Kubernetes/windows/debug/collectlogs.ps1
```diff
@@ -38,14 +39,49 @@ md $outDir
...
+    # V2 Networks
+    Get-HnsNetwork -Version 2 `
+    | Select Name, Type, Id,`
+    @{Name="AddressPrefix"; Expression={$_.Ipams.Subnets.IpAddressPrefix}}` > network.txt
+
+    # V2 Endpoints
+    Get-HnsEndpoint -Version 2 `
+    | Select @{Name="IpAddress"; Expression={$_.IpConfigurations.IpAddress}}, `
+    MacAddress, `
+    (@{Name="IsLocal"; Expression={$_.Health.Extra.Resources.Allocators.IsLocal | Where-Object {$_ -ne $null}}}), `
```
daschott, line 65: This field does not get populated if v1 HNS endpoints were created with Azure CNI on Docker. Wondering if we should just use the V1 equivalent (negated) field "IsRemoteEndpoint" in all cases until official v1.23 deprecation.

daschott: I will be using the v1 IsRemoteEndpoint instead.

daschott:
Validated this on both docker 1.21.9 and containerD v1.24 clusters on AKS.

PR #538, opened 2022-11-07
fix IP concatenation bug

vlappenbusch:
Some IPs to skip were being concatenated by string value, which could have been caused by initially setting the $ips variable to a single IP with the get-pacamapping call

### DIFF
--- a/SDNExpress/scripts/SDNExpressModule.psm1
+++ b/SDNExpress/scripts/SDNExpressModule.psm1
@@ -2206,15 +2206,13 @@ function Initialize-SDNExpressGateway {
         #need to find an address that is not in use
         $ips = @()
 
-        #1 - get-pacamapping from a host to find PAs
+        #1 - get-pacamapping from a host to find PAs and add to list of IPs to not use
         $PACAIps = invoke-command -computername $hostname {
             ((get-pacamapping) | select-object 'PA IP Address').'PA IP Address'
         }
 
         if ($null -ne $PACAIps) {
             $ips += $PACAIps
-        } else {
-            Write-LogInfo $OperationID "PA IPs is null. Setting to empty array"
         }
 
         #2 - get network interfaces and reservations from subnet so we know which IPs to skip over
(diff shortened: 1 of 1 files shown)


vlappenbusch:
@microsoft-github-policy-service agree [company="Microsoft"]

vlappenbusch:
@microsoft-github-policy-service agree company="Microsoft"

SDNExpress/scripts/SDNExpressModule.psm1
```diff
@@ -2207,17 +2207,31 @@ function Initialize-SDNExpressGateway {
...
-        $IPs = invoke-command -computername $hostname {
+        $PACAIps = invoke-command -computername $hostname {
             ((get-pacamapping) | select-object 'PA IP Address').'PA IP Address'
         }
 
-        #2 - get network interfaces from subnet
+        if ($null -ne $PACAIps) {
+            $ips += $PACAIps
+        } else {
+            Write-LogInfo $OperationID "PA IPs is null. Setting to empty array"
```
sbgms, line 2217: It doesnt look like we're setting it. Maybe we should skip the trace ?

SDNExpress/scripts/SDNExpressModule.psm1
```diff
@@ -2207,17 +2207,31 @@ function Initialize-SDNExpressGateway {
...
+
+        #2 - get network interfaces and reservations from subnet so we know which IPs to skip over
         foreach ($ipconfig in $logicalsubnet.properties.ipconfigurations) {
             $ipconfig = get-networkcontrollernetworkinterfaceipconfiguration -connectionuri $uri @CredentialParam -NetworkInterfaceId $ipconfig.resourceRef.split('/')[2] -ResourceId $ipconfig.resourceRef.split('/')[4]
             if (![string]::IsNullOrEmpty($ipconfig.properties.privateipaddress)) {
                 $ips += $ipconfig.properties.privateipaddress
             }
         }
+
+        foreach ($res in $logicalsubnet.Properties.IpReservations) {
```
sbgms, line 2228: Add a comment why we're doing it. Someone might need to know this in future.

sbgms (changes requested):

PR #550, opened 2022-12-22
NC Node Names, Gen1 VMS, Timeout

wu-lily:
Add Network Controller Node Names to Registry Keys so WAC users don't have to manually input this information Add Support for Generation 1 VMs based on whether a VHD was supplied Increase the timeout from 20 minutes to 30 minutes https://microsoft.visualstudio.com/SME/_git/msft-sme-sdn-management/pullrequest/7962923


sbgms (approved):
Thanks for the change!

vlappenbusch (approved):

PR #568, opened 2023-03-25
dump counter values of vfp rules

jayanthaMicrosoft:
Adding an additional "/get-rule-counter" to list of commands used to dump VFP data for a VFP-Port.


Kubernetes/windows/debug/dumpVfpPolicies.ps1
```diff
@@ -37,6 +37,7 @@ foreach ($port in $ports) {
 	& $vfpCtrlExe /list-rule  /port $portGuid | Out-File $outfile -Encoding ascii -Append
```
debj1t, line 37: can you replace the /list-rule command with /get-rule-counter instead of appending it?
/get-rule-counter = /list-rule with rule counters

debj1t (changes requested):

debj1t (approved):

princepereira (approved):
lgtm

daschott (commented):
Why /list-rule is removed? Shouldn't we add /get-rule-counter and preserve /list-rule?

PR #571, opened 2023-04-27
Enabling windows flags for performance improvement.

princepereira:
Fixing bad performance when updating HNS policies for services in Windows nodes.

Changes needed in windows nodes.

1.	Add registry key to enable the feature:
       PS> reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement\Overrides /v 3444861581 /t REG_DWORD /d 1 /f
2.	Update group policy:
        PS> gpupdate /force
3.	Run tasks to read the changes in #1:
        PS> schtasks /run /tn "Microsoft\Windows\Flighting\FeatureConfig\ReconcileFeatures"
4.	Restart computer for changes to take effect (one of the below)
a.	Via cmd: Shutdown -r -t 0
b.	Via PowerShell: Restart-Computer

### DIFF
--- a/Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
+++ b/Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
@@ -1,0 +1,66 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: enable-win-flags
+  labels:
+    app: enable-win-flags
+spec:
+  selector:
+    matchLabels:
+      name: enable-win-flags
+  template:
+    metadata:
+      labels:
+        name: enable-win-flags
+    spec:
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM"
+      hostNetwork: true
+      containers:
+      - name: enable-win-flags
+        image: mcr.microsoft.com/dotnet/framework/samples:aspnetapp
+        command:
+        - powershell.exe
+        - -command
+        - |
+            function EnableWinFlags {
+
+              $count = (reg query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement\Overrides | sls "REG_DWORD" ).Count
+              $regKeySet = (reg query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement\Overrides | sls "3444861581" ).Count
+
+              if(($count -ge 3) -and ($regKeySet -EQ 1)) {
+                Write-Host "Win flags are already enabled."
+                return
+              }
+
+              Write-Host "Adding registry key to enable the feature"
+              reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement\Overrides /v 3444861581 /t REG_DWORD /d 1 /f
+              Start-Sleep -Seconds 2
+
+              Write-Host "Updating group policy"
+              gpupdate /force
+              Start-Sleep -Seconds 2
+
+              Write-Host "Running tasks to read the changes in step 1"
+              schtasks /run /tn "Microsoft\Windows\Flighting\FeatureConfig\ReconcileFeatures"
+              Start-Sleep -Seconds 2
+
+              Write-Host "Restarting computer for changes to take effect"
+              Start-Sleep -Seconds 5
+              Restart-Computer
+            }
+
+            EnableWinFlags
+
+            While($true) {
+              Write-Host "Ready to delete host process container."
+              Start-Sleep -Seconds 60
+            }
+
+        securityContext:
+          privileged: true
+      nodeSelector:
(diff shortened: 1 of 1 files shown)


Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
```diff
@@ -0,0 +1,66 @@
...
+        name: enable-win-flags
+    spec:
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM"
+      hostNetwork: true
+      containers:
+      - name: enable-win-flags
+        image: mcr.microsoft.com/dotnet/framework/samples:aspnetapp
```
daschott, line 23: Curious if we could use this image instead:
`mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0`

https://github.com/microsoft/windows-host-process-containers-base-image#dockerfile-example

It runs on both WS2022 and WS2019 and it is a very small image size.

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
```diff
@@ -0,0 +1,66 @@
...
+            While($true) {
+              Write-Host "Ready to delete host process container."
+              Start-Sleep -Seconds 60
+            }
+
+        securityContext:
+          privileged: true
+      nodeSelector:
+        kubernetes.azure.com/os-sku: Windows2022
+        kubernetes.io/hostname: aksnpwin000000
```
daschott, line 66: Do we want to run this DaemonSet on every Node, or only on one node?
kubernetes.io/hostname: aksnpwin000000 Would mean we schedule it only on the Node called `aksnpwin000000` 
Some customers may have different name for their nodes.

If we want to schedule on only specific nodes, it might be better to use tag instead of hostname. Then they will not need to edit the manifest.

Instead we ask user to add tag to the nodes where they want fix applied using something like
`kubectl label nodes <node-1-name> winflags=3444861581`

And we replace `kubernetes.io/hostname: aksnpwin000000` with filter for our label `winflags:3444861581`

daschott (commented):
LGTM  this should work but I added some potential suggestions

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
princepereira: Thanks David. I will try with this image.

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
princepereira: @daschott If the user wants to update per cluster, will daemonset be the teh right approach ?

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
princepereira: @daschott Tried suggested image, but got following error.

 Normal   Pulling    12s (x2 over 27s)  kubelet            Pulling image "mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0"
  Warning  Failed     12s (x2 over 27s)  kubelet            Failed to pull image **"mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0": rpc error: code = NotFound desc = failed to pull and unpack image "mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0": no match for platform in manifest: not found**
  Warning  Failed     12s (x2 over 27s)  kubelet            Error: ErrImagePull

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
daschott: Yes, if you want to run it on all the Windows Nodes in the cluster, just remove the `kubernetes.io/hostname: aksnpwin000000`. 
Then it will target all Windows nodes and you don't need to add any labels.

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
daschott: image
strange it seemed to have pulled for me on fresh Win2022 AKS Node.... I guess you can leave it as-is.

Kubernetes/windows/debug/enablewinflags/winserver2022.yaml
daschott: And yes, daemonset is the right approach

princepereira: @daschott  Shall I leave the decision to the customer. Because they might have different approach. I will convey the same to them to choose what is best for them.

daschott: sounds good

daschott:
LGTM, lmk if there are any more changes pending otherwise I will merge this on Monday PST

princepereira:
Thanks @daschott

PR #578, opened 2023-06-05
Script to identify network issues. As of now, it looks for DNS issues. It can be extended to capture other network issues as well.

princepereira:
…. It can be extended to capture other network issues as well.

### DIFF
--- a/Kubernetes/windows/startNetworkDiagnostics.ps1
+++ b/Kubernetes/windows/startNetworkDiagnostics.ps1
@@ -1,0 +1,255 @@
+param (
+    [Parameter(Mandatory=$false)]
+    [bool]$DnsOnly = $false,
+
+    [Parameter(Mandatory=$false)]
+    [bool]$DnsPktCap = $false
+)
+
+$Global:All = $true
+
+if ($DnsOnly -eq $true) {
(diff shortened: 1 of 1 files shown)


Kubernetes/windows/startNetworkDiagnostics.ps1
```diff
@@ -0,0 +1,255 @@
...
+    return $freeRangesCount
+}
+
+
+function CheckHnsDnsRuleMissing {
+    $expectedDnsRuleCount = 2
+    Write-Host "Checking HNS DNS Rule missing"
+    $dnsRuleCount = ((Get-HnsPolicyList).Policies | where InternalPort -EQ 53 | where ExternalPort -EQ 53).Count
+    if($dnsRuleCount -lt $expectedDnsRuleCount) {
+        Write-Host "HNS DNS rule count is $dnsRuleCount. DNS issue for sure." -ForegroundColor Red
```
jayanthaMicrosoft, line 91: Instead of conclusive statements like "issue for sure", I think it's better to use statements which only mention the state. In this print statement, it's enough to mention "HNS DNS rule count is $dnsRuleCount" (in my opinion).

PR #585, opened 2023-11-22
Backport sdnexpress from sdn tools to github repository

sbgms:
Backported the following fixes/features:

- support to deploy using rest IP instead of rest name
- prefer certificates with nc oid in their ekus instead of server & client oids
- (preview) of deployment over a failover cluster


nijosmsft (approved):

sbgms:
Thanks!

